PR #6, opened 2022-03-05
to add polkadot.md

bumblefudge:
I totally went out on a limb and defined CAIP-10 for this because it seemed so well-documented and straight-forward, although I had some questions:

1. Would it be clearer or more standard/intuitive for Polkadot developers to use the prefix registered in the SS58 directory as the chainID? if nothing else, it would shorten CAIP-10s
2. Even shorter would just be removing the chain reference from the CAIP-10s, since they are recoverable from the addresses! Is that... even an option, without updating CAIP-10?  If so, would they look like
`polkadot:EVH78gP5ATklKjHonVpxM8c1W6rWPKn5cAS14fXn4Ry5NxK`
or
`polkadot::EVH78gP5ATklKjHonVpxM8c1W6rWPKn5cAS14fXn4Ry5NxK`
?
3. I didn't even touch NFTs or assets, but maybe I should open a tracking issue for it?

CC @pedrouid @joshua-mir @shawntabrizi

### DIFF
--- a/polkadot/caip2.md
+++ b/polkadot/caip2.md
@@ -11,20 +11,28 @@ requires: CAIP-2
 supersedes: CAIP-13
 ---
 
-# Context
+# CAIP-2
 
 *For context, see the [CAIP-2][] specification.*
+
+## Rationale
 
 The namespace is called "polkadot" to refer to Polkadot-like "parachains,"
 parallel and independent Layer 1s coordinated by the Polkadot relay chains.
 
-## Reference Definition
+The rationale behind the use of block hash from the genesis block stems from its
+usage in the Polkadot architecture in network and consensus.
+
+## Syntax
 
 The definition for this namespace will use the `genesis-hash` as an indentifier
 for different Polkadot chains. The format is a 32 character prefix of the block
 hash (lower case SHA256 hex).
 
-## Resolution Method
+As CAIP-2 references for this namespace are 32-byte SHA256 hashes in lowercase
+hex, they can be validated with the simple regular expression: `[0-9a-f]{32}`
+
+### Resolution Method
 
 To resolve a blockchain reference for the Polkadot namespace, make a JSON-RPC
 request to the blockchain node with method `chain_getBlockHash`, for example:
@@ -49,21 +57,11 @@ The response will return as a value for the result a hash for the block with
 height 0 that should be sliced to its first 16 bytes (32 characters for base 16)
 to be CAIP-2/CAIP-10 compatible.
 
-## Rationale
-
-The rationale behind the use of block hash from the genesis block stems from its
-usage in the Polkadot architecture in network and consensus.
-
-## Syntax
-
-As CAIP-2 references for this namespace are 32-byte SHA256 hashes in lowercase
-hex, they can be validated with the simple regular expression: `[0-9a-f]{32}`
-
 ### Backwards Compatibility
 
 Not applicable
 
-### Test Cases
+## Test Cases
 
 This is a list of manually composed examples
 
--- a/polkadot/caip2.md
+++ b/polkadot/caip2.md
@@ -17,8 +17,10 @@ replaces: CAIP-13
 
 ## Rationale
 
-The namespace is called "polkadot" to refer to Polkadot-like "parachains,"
-parallel and independent Layer 1s coordinated by the Polkadot relay chains.
+The namespace is called "polkadot" to encompass all chains in the polkadot
+ecosystem, allowing actors and entities to be addressable from core Polkadot
+"relay" chains, indepedent "solo" chains, and L2-like "parachains" coordinated
+by the Polkadot relay chains. 
 
 The rationale behind the use of block hash from the genesis block stems from its
 usage in the Polkadot architecture in network and consensus.
@@ -85,6 +87,7 @@ polkadot:37e1f8125397a98630013a4dff89b54c
 - [Polkadot address explainer][]: A quick overview of how network-specific,
       self-describing addresses can derive from the same private key
 - [Polkadot subscan tool][]: A tool for transforming addresses according to SS58 across polkadot networks
+- 
 
 [Polkadot address explainer]: https://www.quora.com/How-do-different-wallet-addresses-work-on-Polkadot-and-Kusama
 [Polkadot identity system]: https://wiki.polkadot.network/docs/learn-identity
--- a/polkadot/caip2.md
+++ b/polkadot/caip2.md
@@ -1,39 +1,30 @@
 ---
-namespace-identifier: polkadot
-title: Polkadot Namespace
+namespace-identifier: polkadot-caip2
+title: Polkadot Namespace - Chains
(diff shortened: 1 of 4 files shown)


bumblefudge:
@Jam10o are you the same Joshua Mir?  Are you interested at all in contributing to this, or conversely, would you prefer I take your name off future contributions?  Feel free to backchannel me on Discord if you'd like to discuss in private, I'm <GH uname>#1488 there.

bumblefudge:
notes from today's meeting:
- polkadot:_: would not break existing reg-exs 
- maybe referring to relay-chain better than not referring to anything?
- return to "caip-10 without caip-2" corner case later, when someone actually raises an issue!

TimDaub:
In today's call you, @bumblefudge, mentioned that polkadot's addressing scheme integrates the chainID into the address and makes it reproducible. Currently, the namespace is suggested as e.g. `polkadot:...`, whereas the logical equivalent to the addressing in the Ethereum namespace is prefixed by the EIP defining the chainID concept `eip155:1:`.

So to reduce redundancy, I'd suggest prefixing the polkadot namespace with the name of the standard document that defines how chainIDs are encoded into polkadot addresses. E.g. if Polkadot as a "Polkadot Improvement Proposal" process with a PID-XY defining how chainIDs are encoded into the address, we'd prefix with `pid-xy:EVH78gP5ATklKjHonVpxM8c1W6rWPKn5cAS14fXn4Ry5NxK`.

At least those were my thoughts from the top of my head. Haven't looked into Polkadot really but found this:

- https://docs.substrate.io/v3/advanced/ss58/

bumblefudge:
good point!  The namespace is entirely governed by ss58 and its registry (which is in the same repo!) so `ss58:-:EVH78gP5ATklKjHonVpxM8c1W6rWPKn5cAS14fXn4Ry5NxK` sounds like a better name!  I'll update my PR and await a confident 👍 or 👎 from someone with polkadot-building experience!

ntn-x2:
@bumblefudge how would this new proposed scheme to replace CAIP-10 would play out with the CAIP-2 definition of a Polkadot-based chain? In that case, we are still using genesis hashes to identify a particular chain, and this, as far as I understand, would break the requirement that a CAIP-10 address is built on top of a CAIP-2 identifier.

polkadot/caip2.md
```diff
@@ -0,0 +1,109 @@
...
+```
+# Kusama
+polkadot:b0a8d493285c2df73290dfb7e61f870f
+
+# Edgeware
+polkadot:742a2ca70c2fda6cee4f8df98d64c4c6
+
+# Kulupu
+polkadot:37e1f8125397a98630013a4dff89b54c
+```
```
joepetrowski, line 77: "The namespace is called "polkadot" to refer to Polkadot-like "parachains,""

None of these are parachains :). Ideally namespaces would anchor at consensus systems, so e.g.:

```
polkadot:<statemint-genesis-hash>
kusama:<statemine-genesis-hash>
```

Or, if you wanted to express that Kusama and Polkadot are connected, perhaps add another "relational" namespace in the path, like:

```
polkadot:canary:<kusama-genesis-hash>
polkadot:para:<statemint-genesis-hash>
polkadot:solo:<kulupu-genesis-hash>
```

The problem that will come up is migrations. For example, Centrifuge and Integritee (maybe some others, these two off the top of my head) were solo chains that became parachains. So if these identifiers are meant to last forever, there could be issues, but I'm really not that familiar with CAIP.

Or, if you want to avoid migration issues entirely, just change the docs from "The namespace is called "polkadot" to refer to Polkadot-like "parachains,"" to something like, "The namespace is called "polkadot" to refer to blockchains whose state transition functions can execute on the Polkadot host API, for example parachains or Kusama."

ntn-x2: In the latter example, would all chains then still use the `polkadot` namespace? I saw a comment about the `MultiLocation` struct that mentions forks also for (para)chains in the Polkadot space. If that is the case, I think having a genesis hash as a way to identify chains is definitely not sufficient, as two forked chains would be identified by the same hash. To this point, would a combination of genesis hash and SS58 prefix be sufficient to uniquely identify all chains, including forks and migrated ones?

joepetrowski: > In the latter example, would all chains then still use the polkadot namespace? 

Yes, as they are chains that execute on top of the Polkadot host API (which is not just expressed in Substrate, but also in other implementations like ChainSafe's Gossamer).

> I saw a comment about the MultiLocation struct that mentions forks also for (para)chains in the Polkadot space. If that is the case, I think having a genesis hash as a way to identify chains (as is the case now) is definitely not sufficient, as two forked chains would be identified by the same hash.

How do you define a fork? A parachain should never fork, unless someone starts a new chain with a modified chainspec/runtime, but then that's just a new genesis. `MultiLocation`s in general are _relative_ location paths but don't actually know about a `/` location (although in practice it would be the governing consensus system). Since CAIP does seem to have a `/` location (the first namespace), `MultiLocation` could be mapped onto it from one chain ID to another.

> would a combination of genesis hash and SS58 prefix be sufficient to uniquely identify all chains, including forks and migrated ones?

No, SS58 should probably not even come near this conversation. It's 99% for front end, there is no guarantee that it's unique (already, Polkadot and Statemint both use a prefix of 0), nor that it will stay the same in a chain's lifetime.

ntn-x2: Sorry, my mistake. Forks could nevertheless still exist for standalone chains, and if they would be identified with the same `polkadot` namespace, then using the sole genesis hash would not be sufficient anymore.

> No, SS58 should probably not even come near this conversation. It's 99% for front end, there is no guarantee that it's unique (already, Polkadot and Statemint both use a prefix of 0), nor that it will stay the same in a chain's lifetime.

Yeah that's what I thought. I wanted to make it clear as there has been conversation up above about using those to identify chains.

bumblefudge: some time next month I'll devote a big bloc of time to overhauling all this and tag you both in the PR-- I clearly need to go back to the drawing board to map the CAIP mental model onto this VM!

ntn-x2:
@bumblefudge is there anything I can do to resume the discussion in this PR and move it forward? I opened [another issue](https://github.com/ChainAgnostic/CAIPs/issues/215) in this repo which somehow seems to be related to having a definition for the Polkadot namespace. Would like some clarity both on this PR and on that issue 😁

bumblefudge:
Hehe, yeah, that's fair.  Two related issues:

1.) polkadot PRs have never seemed to get quorum/consensus because I don't understand polkadot from shinola and all attempts at reading enough to merge this PR end in my stuck in a rabbithole not understanding coordination chains or polkadot dev community needs.  in particular, what's confusing to me is the difference between chainID systems and polkadot's system for addressing sidechains and private chains. 

2.) As of last week, you are NO LONGER the only dev who has requested a CAIP for on-chain attestations.  Let's try to merge a namespace for polkadot first (for my benefit, literally!) so that we can make sure the attestation CAIP we right after that makes sense across diff types of chains!

polkadot/caip2.md
bumblefudge: Going to merge a straw man and invite PRs to improve, will return to this discussion

bumblefudge:
Discussed out-of-band with @ntn-x2 -- Merging as-is to allow an overhaul in ongoing issues and PRs.

PR #20, opened 2022-03-27
to extend polkadot/caip10 to include "implicit/not-yet-derived" CAIP-2

bumblefudge:
I realized that "polkadot:-:{address}" scheme would break CAIP-2 because `-` is no more a valid `chainID` than ` ` or null would be; then I re-read [Tim's comment](https://github.com/ChainAgnostic/namespaces/pull/6#issuecomment-1064299030) and realized I had misinterpreted it! 

In any case, I'm still not sure this complexity is justified by real polkadot usage so I'm going to leave this PR open until I hear from people that have experience of that ecosystem and its practices.

### DIFF
--- a/polkadot.md
+++ b/polkadot.md
@@ -1,0 +1,162 @@
+---
+namespace-identifier: polkadot
+title: Polkadot Namespace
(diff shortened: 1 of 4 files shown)


bumblefudge:
in terms of wildcards and non-chain IDs in the chainId segment of CAIP-10, there is now precedent in another namespace that may be relevant to this PR:
https://github.com/ChainAgnostic/namespaces/pull/107

PR #25, opened 2022-06-01
extends solana/caip19

qbig:
---
namespace-identifier: solana-caip19
title: Solana Namespace - Asset
author: Qiao Liang (@qbig)
discussions-to: 
status: Draft
type: Standard
created: 2022-06-01
updated: 2022-06-01
requires: ["CAIP-2", "CAIP-10", "CAIP-19", "CAIP-30"]
---

# CAIP-19

*For context, see the [CAIP-19][] specification.*

## Rationale

On solana, all contracts are stateless and state is instead represented using "accounts". More specifically, all fungible and non-fungible tokens are using the same instance of token contracts and for each token:

- 1 single global "mint" account instance is used for the global data like "total supply".
- each token account represent an account balance belong to a certain address.

And the difference between fungible token and non-fungible token is that non-fungible token mint has a total supply of 1 and zero decimal place.
But both asset type can be identified by their "mint" account. 
qbig marked this conversation as resolved.
Show resolved

## Syntax

After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `token`, `nft` as the namespace and the address of the `mint` account as the reference.

| Reference   | Equivalent to | See also            |
| :---        | :----         | :---                |
| token       | ERC20 Token   | [Metaplex](https://docs.metaplex.com/programs/token-metadata/token-standard) Fungible   |
| nft         | ERC721        | [Metaplex](https://docs.metaplex.com/programs/token-metadata/token-standard) NonFungible|

## Examples

```
# One Solana Mainnet NFT：Pesky Penguins #398
solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/token:Fz6LxeUg5qjesYX3BdmtTwyyzBtMxk644XiTqU5W3w9w
# Solana Mainnet USDC
solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/token:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
```

## References

[Address Lookup Table Proposal]: https://docs.solana.com/proposals/transactions-v2
[Account Types]: https://docs.solana.com/terminology#account
[Address Expressions]: https://docs.solana.com/cli/transfer-tokens#receive-tokens
[Token Mint]: https://spl.solana.com/token#creating-a-new-token-type
[CAIP-2]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md
[CAIP-10]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md
[CAIP-19]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md
[CAIP-30]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-30.md

## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

### DIFF
--- a/solana/caip19.md
+++ b/solana/caip19.md
@@ -27,17 +27,17 @@ But both asset type can be identified by their "mint" account.
 
 ## Syntax
 
-After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `spl-token` as the namespace and the address of the mint account as the reference.
+After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `token` as the namespace and the address of the mint account as the reference.
 
 
 ## Examples
 
 ```
 # One Solana Mainnet NFT：Pesky Penguins #398
-solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/spl-token:Fz6LxeUg5qjesYX3BdmtTwyyzBtMxk644XiTqU5W3w9w
+solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/token:Fz6LxeUg5qjesYX3BdmtTwyyzBtMxk644XiTqU5W3w9w
 
 # Solana Mainnet USDC
-solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/spl-token:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
+solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/token:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
 ```
 
 ## References
--- a/solana/caip19.md
+++ b/solana/caip19.md
@@ -27,7 +27,12 @@ But both asset type can be identified by their "mint" account.
 
 ## Syntax
 
-After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `token` as the namespace and the address of the mint account as the reference.
+After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `token`, `nft` as the namespace and the address of the `mint` account as the reference.
+
+| Reference   | Equivalent to | See also            |
+| :---        | :----         | :---                |
+| token       | ERC20 Token   | [Metaplex](https://docs.metaplex.com/programs/token-metadata/token-standard) Fungible   |
+| nft         | ERC721        | [Metaplex](https://docs.metaplex.com/programs/token-metadata/token-standard) NonFungible|
 
 
 ## Examples
--- a/solana/caip19.md
+++ b/solana/caip19.md
@@ -1,0 +1,70 @@
+---
+namespace-identifier: solana-caip19
+title: Solana Namespace - Asset
+author: Qiao Liang (@qbig)
+discussions-to: 
+status: Draft
+type: Standard
+created: 2022-06-01
+updated: 2022-06-01
+requires: ["CAIP-2", "CAIP-10", "CAIP-19", "CAIP-30"]
+---
+
+# CAIP-19
+
+*For context, see the [CAIP-19][] specification.*
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


solana/caip19.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Rationale
+
+On solana, all contracts are stateless and state is instead represented using "accounts". More specifically, all fungible and non-fungible tokens are using the same instance of token contracts and for each token:
+
+- 1 single global "mint" account instance is used for the global data like "total supply".
+- each token account represent an account balance belong to a certain address.
+
+And the difference between fungible token and non-fungible token is that non-fungible token mint has a total supply of 1 and zero decimal place.
+But both asset type can be identified by their "mint" account.
```
bumblefudge, line 25: This is great, thanks for including this.

solana/caip19.md
```diff
@@ -0,0 +1,54 @@
...
+- 1 single global "mint" account instance is used for the global data like "total supply".
+- each token account represent an account balance belong to a certain address.
+
+And the difference between fungible token and non-fungible token is that non-fungible token mint has a total supply of 1 and zero decimal place.
+But both asset type can be identified by their "mint" account. 
+
+
+## Syntax
+
+After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `spl-token-mint` as the namespace and the address of the mint account as the reference.
```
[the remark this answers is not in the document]
qbig: I think "mint" is probably the clearest option here, considering both Solana and non-solana devs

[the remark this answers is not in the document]
qbig: or `spl-token` since there might be non-spl-token in future.

qbig:
updated using `spl-token` as asset namespace

qbig:
@bumblefudge could you take another look?

qbig:
updated using `token` as asset namespace

solana/caip19.md
```diff
@@ -0,0 +1,57 @@
...
+- 1 single global "mint" account instance is used for the global data like "total supply".
+- each token account represent an account balance belong to a certain address.
+
+And the difference between fungible token and non-fungible token is that non-fungible token mint has a total supply of 1 and zero decimal place.
+But both asset type can be identified by their "mint" account. 
+
+
+## Syntax
+
+After the CAIP-2 namespace+chainID, a slash defines an `asset_namespace` and an `asset_reference`. And since both fungible and non-fungible tokens can be identified using the mint account, we could use `token` as the namespace and the address of the mint account as the reference.
```
felipe-chamas, line 30: How about Metaplex [Token Metadata v1.1](https://docs.metaplex.com/programs/token-metadata/changelog/v1.1) which now includes [Token Standard Field](https://docs.metaplex.com/programs/token-metadata/token-standard#the-token-standard-field)?
In this sense we could use:
```ts
// Equivalent to ERC20
'solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/Fungible:EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v'

// Equivalent to ERC721
'solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/NonFungible:Fz6LxeUg5qjesYX3BdmtTwyyzBtMxk644XiTqU5W3w9w'

// Equivalent to ERC1155
'solana:4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZ/NonFungibleEdition:Fz6LxeUg5qjesYX3BdmtTwyyzBtMxk644XiTqU5W3w9w'
```

felipe-chamas (commented):
I think it'd also be nice if you could include the CAIP-22 proposal. I think your issue is titled with CAIP-19, but the ERC721 example with TokenId is based on CAIP-22

solana/caip19.md
qbig: @felipe-chamas it's a great suggestion as it indeed implies more information. but two issues with Token Standard Field is that they are:
1. long
2. implied rather than speficied directly (if this and this, then it's NonFungibleEdition)

So how about just `token` and `nft`, `edition` ?

qbig: @bumblefudge @ligi tht?

bumblefudge: yes i prefer the abbreviations, maybe just add a little table like

|reference|equivalent to|see also|
|---|---|---|
| `token` | native tokens, equivalent | [Metaplex](https://docs.metaplex.com/programs/token-metadata/token-standard#the-token-standard-field) `Fungible` |
...
?

obstropolos:
@qbig - we can move to get this in - can you add a quick description to the PR as well?

qbig:
@obstropolos could you take a look when it's convenient?

bumblefudge (approved):

PR #26, opened 2022-06-08
add solana caip10

haardikk21:


oed (commented):
Could you add a list of chain name -> hash used as chain_id?

haardikk21:
> 

Added it, with a reference to CAIP-2.

bumblefudge (approved):
LGTM

PR #31, opened 2022-08-26
solana: Change CAIP-122 spec to be similar to eip155 one

ukstv:
Apparently, we messed with ordering of the fields. ChainID should be higher in the stringified representation of signing input. Now it is the same as for eip155 aka EIP-4361.


oed (approved):
Rugpulled by @spruceId :D

ligi (approved):

PR #33, opened 2022-09-01
to add arweave/README + caip2 + caip10 + caip122

ropats16:
Create a new namespace for arweave. 

Add support for:
- caip2 (Chains)
- caip10 (Addresses)
- caip122 (SIWX)

### DIFF
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -22,7 +22,7 @@ Particularities of syntax for Arweave "accounts" have been specified.
 
 ## Syntax
 
-Arweave addresses have a normalization of lowercase letters, uppercase letters, numbers and `-`s.
+Arweave addresses have a normalization of lowercase letters, uppercase letters, numbers and `-`s. The addresses are encoded in base64Url format.
 
 ### Backwards Compatibility
 
@@ -32,7 +32,7 @@ Not applicable.
 
 ```
 # Arweave mainnet
-ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
+arweave:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ```
 
 ## References
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -41,7 +41,7 @@ ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 - [Arweave Transaction Signing](https://docs.arweave.org/developers/server/http-api#transaction-signing)
 - [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
 - [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
-- [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md): Sign in With X 
+- [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md)
 
 
 
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -38,8 +38,10 @@ ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ## References
 
 - [Arweave](https://github.com/ArweaveTeam/arweave-standards)
+- [Arweave Transaction Signing](https://docs.arweave.org/developers/server/http-api#transaction-signing)
 - [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
 - [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
+- [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md): Sign in With X 
 
 
 
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -1,7 +1,7 @@
 ---
 namespace-identifier: arweave-caip10
 title: Arweave Namespace - Addresses
-author: Rohit Pathare (@ropats16), Phil Billingsby (@pbillingsby)
+author: Rohit Pathare (@ropats16), Phil Billingsby (@pbillingsby), Dan MacDonald (@DanMacDonald)
 discussions-to:
 status: Draft
 type: Standard
@@ -32,12 +32,13 @@ Not applicable.
 
 ```
 # Arweave mainnet
-arweave:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
+arweave:7wIU7KolICAjClMl:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ```
 
 ## References
 
 - [Arweave](https://github.com/ArweaveTeam/arweave-standards)
+- [Hash of Genesis Block](https://viewblock.io/arweave/block/0)
 - [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
 - [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
 - [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md)
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -32,7 +32,7 @@ Not applicable.
 
 ```
 # Arweave mainnet
-arweave:7wIU7KolICAjClMl:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
+arweave:7wIU:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ```
 
 ## References
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -38,7 +38,6 @@ ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ## References
 
 - [Arweave](https://github.com/ArweaveTeam/arweave-standards)
-- [Arweave Transaction Signing](https://docs.arweave.org/developers/server/http-api#transaction-signing)
 - [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
 - [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
 - [CAIP-122](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-122.md)
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -38,8 +38,8 @@ ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
 ## References
 
 - [Arweave](https://github.com/ArweaveTeam/arweave-standards)
-- [CAIP-2]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md
-- [CAIP-10]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md
+- [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
+- [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md)
 
 
 
--- a/arweave/caip10.md
+++ b/arweave/caip10.md
@@ -1,0 +1,48 @@
+---
+namespace-identifier: arweave-caip10
+title: Arweave Namespace - Addresses
+author: Rohit Pathare (@ropats16), Phil Billingsby (@pbillingsby)
+discussions-to:
+status: Draft
+type: Standard
+created: 2022-09-01
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+*For context, see the [CAIP-10][] specification.*
+
+## Abstract
+In CAIP-10 an account identification scheme is defined. This is the implementation of CAIP-10 for Arweave network.
+
+## Rationale
+
+Particularities of syntax for Arweave "accounts" have been specified.  
+
+## Syntax
+
+Arweave addresses have a normalization of lowercase letters, uppercase letters, numbers and `-`s.
+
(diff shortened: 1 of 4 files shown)


ligi:
are you in Berlin at Berlin blockchain week? Our next meeting to discuss PRs is in person there - see: https://github.com/ChainAgnostic/Gatherings

ropats16:
> 

I won't be able to join the Berlin blockchain week but I'd like to know the consensus y'all reach and how we can take this further.

arweave/caip10.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Backwards Compatibility
+
+Not applicable.
+
+## Test Cases
+
+```
+# Arweave mainnet
+ar:1:kY9RAgTJEImkBpiKgVeXrsGV02T-D4dI3ZvSpnn7HSk
```
oed, line 35: Can you mention somewhere what format the address is encoded with? Is it base64?

oed (commented):
Looks mostly good. Would like @zachferland to have a look at the caip122 spec

arweave/caip10.md
DanMacDonald: base64Url encoded 😎

arweave/caip2.md
```diff
@@ -0,0 +1,56 @@
...
+
+Not applicable.
+
+## Test Cases
+
+This is a manually composed example.
+
+```
+# Arweave Mainnet
+arweave:1
```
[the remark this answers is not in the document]
ropats16: Arweave has only 1 chain currently which has been designated `1` in the ecosystem. If more chains are added, a system for identifying chains would be created at the time and the `arweave` namespace would be updated accordingly.

arweave/caip10.md
ropats16: Updated the PR to reflect the format in which the addresses are encoded.

arweave/caip122.md
```diff
@@ -0,0 +1,71 @@
...
+requires: ["CAIP-122", "CAIP-2", "CAIP-10"]
+---
+
+## CAIP-122
+
+For context, see the [CAIP-122](CAIP-122) specification.
+
+## Rationale
+
+On Arweave, [EIP-4361](EIP-4361) Sign-in with Arweave already exists, which was the inspiration for CAIP-122. This specification highlights how EIP-4361 conforms with CAIP-122.
```
zachferland, line 18: does a sign in with arweave standard already exist? this references eth eip

arweave/caip122.md
```diff
@@ -0,0 +1,71 @@
...
+
+The abstract data model must be converted into a string representation in an unambigious format. We use the format as defined in EIP-4361 as reference.
+
+An informal template of the full message is presented below. The field descriptions are provided in the order they must appear as follows:
+- `domain` is the RFC 3986 authority that is requesting the signing.
+- `address` is the Arweave address performing the signing.
+- `uri` is an RFC 3986 URI referring to the resource that is the subject of the signing (as in the subject of a claim).
+- `version` is the current version of the message, which MUST be 1 for this specification.
+- `type` of the signature to be generated, as defined in the namespaces for this CAIP.
+- `signature` is the result of concatenation and signing of the other field descriptions by the wallet.
```
zachferland, line 40: can it additionally include all field also in EIP-4361?

arweave/caip122.md
ropats16: Arweave has not needed a sign in standard in the past because of the way the ecosystem works. However, with this integration, there is a use for it. So this CAIP is the first version of the standard. Would love some insights on how we can rephrase the rationale based on this information.

arweave/caip122.md
ropats16: As mentioned in the comment above, this CAIP is a first version of the standard for signing in with arweave and we are only using the EIP-4361 as a reference. Can having only the required fields for now suffice? We can add more fields as the standard becomes refined.

zachferland: ok, yeah in this case would include all required and optional fields, as most optional are needed to enable full use and options of CACAO, and we (ceramic) expect some optional fields to enable our use cases (ie resources, expiration), others may as well (ie nonce for authn)

arweave/caip122.md
zachferland: yup makes sense, could usual similar wording to here - https://github.com/ChainAgnostic/namespaces/blob/main/solana/caip122.md and mention inspired by eip-4316, or reference

zachferland:
Lastly in the signature creation section, would have to describe how to go from that message to signature creation. Ie from message to bytes/encoding and if arweave has a standard signing interface/implementation (or if ambiguous, add detail). Again ref solana one, which provides a sufficient explanation for example, or eth which has existing EIPs which describe signature creation and verification and are referenced instead.

arweave/caip122.md
```diff
@@ -71,6 +71,8 @@ Type: ${type}
 Signature: ${signature}
 ```
 
+The signature creation process on Arweave is done through the `arweave-js` library. The data model is JSON serialized to a string. This string is signed using the RSA-PSS algorithm through a manual call to the `sign()` method from the `arweave-js` library. The signature is returned as a Uint8Array of bytes which is then Base64Url encoded by calling `arweave.utils.bufferTob64Url()`.
```
zachferland, line 74: message above will already be a string, would you then be able to directly sign that after or would you pass a Uint8Array of the string to be signed?

ropats16: If the data model is already a JSON serialised string then we can directly pass it to the sign() method, getting the Uint8Array which would be encoded to Base64 URL format to get the final signature which is the format that is needed to be sent over HTTP to the server for verification.

ropats16: @zachferland updated the PR to reflect the data model being a string.

zachferland: thanks! im good with the caip122 part 👍

zachferland (approved):

arweave/caip122.md
ropats16: @oed if CAIP122 was the last part we needed to get sorted, can we get this merged so that we can move onto the next steps?

ropats16:
> @oed if you have any more feedback and/or want to approve

@oed following up on this to check if it is good to be merged?

arweave/caip122.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Specification
+
+### Signing Algorithm
+
+Arweave wallets are RSA key pairs and use the private and public keys for signing and verifying signatures. Arweave wallet addresses are the SHA-256 hash of the RSA public key.
+
+### Signature Type
+
+Arweave signatures use the RSA-PSS signing algorithm to sign data and verify signatures.
```
oed, line 28: You need to specify a singature type string. See: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-74.md#signature-verification

arweave/caip2.md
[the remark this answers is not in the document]
oed: There is no testnet at all? Can you also include a reference to where it's designated `1` in the ecosystem?

oed (changes requested):

arweave/caip122.md
ropats16: Do you mean something like Solana has specified on this line here?

https://github.com/ceramicnetwork/cacao/blob/d18d3f40bddd07d7a4ee5959f003c3416c9e1442/src/cacao.ts#L38

arweave/caip2.md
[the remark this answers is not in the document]
ropats16: There is no official test net yet. The designation of the main network as 1 is a value referenced from the EVM compatible ecosystem. As this kind of standard was not needed before this particular use case, there is no document I can point to for the same. This is essentially the first iteration of such a standard for the arweave ecosystem.

arweave/caip122.md
oed: Yes exactly.

arweave/caip2.md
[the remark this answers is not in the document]
oed: In ethereum the number has a specific meaning and all network id's are maintained on a list: https://chainlist.org/

Are there unofficial test nets? If yes, how does the client software differentiate between them?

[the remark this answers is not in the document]
ropats16: The Arweave protocol doesn't have built in support for multiple networks at the protocol level. HTTPS request are made to nodes but there's a single official Arweave network so there isn't a need to get a response from the nodes as to which network they are on. In case this is revised in the future, adequate documentation and PRs for the same would be submitted.

ropats16:
@oed the latest updates specifies the type of the signature as it would be represented in the cacao block. Is this what you were looking for? And if so are there any other changes?

arweave/caip2.md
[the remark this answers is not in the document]
oed: From a quick google it seems like there are indeed testnets. On a protocol level they need to be distinguished. Also if we arbitrarily pick a number here it will fall on CASA to maintain a list of numbers for arweave chains which we definitely don't want to do!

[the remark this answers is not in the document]
DanMacDonald: There are testnets but they aren't actually networks of nodes so there's no need to distinguish them at a the protocol level. They aren't gossping data/transactions  on an open network.  These "testnets" expose Arweave like HTTP endpoints but are hosted as a standalone deployments.  When a developer wants to test their application they point their code at  the URL provided by the testnet and communicate directly with it. Testing works like this on arweave, in part, because the arweave protocol does not have a way to differentiate "network" at the protocol level. 

Development on Arweave is typically done using https://github.com/textury/arlocal, a shim that mocks up the arweave interfaces and mimics its behavior. The "testnets" that exist are a slightly modified version of this code that enable incoming connections to be managed by a load balancer. Example of this would be Redstone's "testnet" https://testnet.redstone.tools/

You may also find references to Testweave-sdk or Testweave-docker. Testweave coexisted with `arlocal` for some time and similarly enabled testing on the developers local machine. Instead of a node/SQLite mockup testweave ran an Arweave node in a container. Arweave node software really wasn't meant to run in this configuration and as a result it had constant bugs and inconsistencies with mainnet. Developers abandoned testweave for the better supported `arlocal` and it is now effectively deprecated with no active maintainer.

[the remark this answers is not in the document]
bumblefudge: is there any canonical "chainID" equivalent for these testnets, though, canonically maintained in any kind of registry on github or otherwise, as Joel was asking? would "testnet.redstone.tools" be a meaningful testnet name, on which someone might want to test a system by addressing and fetching state?

this information looks pretty useful to first-time arweave devs, would it make sense to put it into the README somewhere as an informational aside?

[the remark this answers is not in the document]
joshbenaron: @DanMacDonald Arweave is releasing a testnet for 2.6. Isn't this going to be its own chain?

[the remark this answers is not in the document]
DanMacDonald: That's true, there is a testnet for 2.6, purely for miners. I'm not sure exactly how they are accomplishing this. My guess is though a whitelist of 2.6 peers.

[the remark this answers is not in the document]
ropats16: Hey @oed @bumblefudge, based on the information provided by @DanMacDonald, how can we move forward? As the testnet for 2.6 isn't released yet, ans it's working mechanism is still vague, is it possible to move forth with considering the existence of only the Arweave main net at the moment and once the test net is released, the chain resolution can be revised and updated accordingly in the namespaces as well?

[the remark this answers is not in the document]
oed: Can we maybe use the genesis has as we do for other chains? That way if in the future Arweave introduces a testnet we don't have to maintain a list of numbers.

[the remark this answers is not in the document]
bumblefudge: this makes sense to me-- calling it `1` or `main` is arbitrary if the protocol doesn't internally have the concept of a chain ID and the community itself doesn't use that convention.  i worry that, as Joel rightly points out, the community might proceed happily for years without testnets or forks, and at such a time as they adopt the concept, they might number from 0 or use strings and call it `main-net` or `mainnet` instead of `main`.  For this reason, I second the recommendation for the "hash of genesis block" approach taken by many other caip-2 profiles since current protocol does seem to have the notion of "block 0" and addressing blocks by hashes.  

from [here](https://viewblock.io/arweave/block/0) I take it that the genesis hash is numbered 0, not 1, and hash-addressable as `7wIU7KolICAjClMlcZ38LZzshhI7xGkm2tDCJR7Wvhe3ESUo2-Z4-y0x1uaglRJE` 
. from [this documentation] i am inferring it that since block 422250 the format of RPC calls changed, but the block hashes still seem at first glance to be computed the same way and would fit the same regex: 
`JbtpfyJ6Kh22vciZJpZAVorejqDRB6ekbZC0QF_zOFbACpcae6L6PZsuy0gKRpqk` . 
It's interesting that the RPC doesn't include a request for block by block number, just by hash, but at least you can confirm you're on the intended chain (or at least, a fork of it...) by calling 
[`GET https://arweave.net/block/hash/7wIU7KolICAjClMlcZ38LZzshhI7xGkm2tDCJR7Wvhe3ESUo2-Z4-y0x1uaglRJE`](https://docs.arweave.org/developers/server/http-api#blocks) and checking the `indep_hash` top-level property of the result object to confirm.  It seems like that flow still works after the info RPC changes.

Anyways if this works for everyone, feel free to update the CAIP-2 profile and we can go from there-- tag me i you'd like any help.  To limit the length of CAIP2 and CAIP10 URNs to fit inside, say, HTTP headers, we have historically just substringed the first 16 or 32 characters-- the full 64 would make for unworkably long CAIP-10s and CAIP-19s.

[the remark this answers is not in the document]
oed: A trunkated hash of genesis might be alright as well. So the caip2 doesn't become so long. Maybe first 4 characters?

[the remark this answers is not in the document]
ropats16: Hey @bumblefudge @oed, just updated the PR to identify the network using a truncated version (first 16 characters) of the [hash of the genesis block](https://viewblock.io/arweave/block/0) in both `CAIP2` and `CAIP10` files.

bumblefudge:
Chain ID stuff looks great-- deferring to the Spruce or Ceramic teams for confirmation that the signature suite stuff is encoded conformantly, will merge after that.

arweave/caip2.md
```diff
@@ -22,7 +22,7 @@ The namespace "arweave" refers to the wider Arweave ecosystem.
 
 #### Reference Definition
 
-The reference relies on Arweave's current designation of addresses belonging to main network by prefixing them with `1`.
+The reference relies on Arweave's current designation of addresses belonging to main network by prefixing them with `7wIU7KolICAjClMl` which is the reprresentation of the [hash of the genesis block](https://viewblock.io/arweave/block/0) truncated to the first 16 characters.
```
oed, line 25: I would make it even shorter! 4 characters should be plenty. Fairly low risk of collision here.

oed (approved):
Alright by me. Made one optional suggestion.

arweave/caip2.md
ropats16: Hey @oed just made the suggested changes.

oed (approved):
lgtm

ropats16:
@oed curious over who's supposed to merge it? Don't think I have the permissions for it.

oed:
It will have to be approved in the next CASA meeting: https://github.com/ChainAgnostic/CASA#regular-meetings

PR #39, opened 2022-11-21
Spec for StarkNet

delaaxe:
Adds CAIPs for StarkNet.

### DIFF
--- a/starknet/README.md
+++ b/starknet/README.md
@@ -14,11 +14,13 @@ replaces: CAIP-28
 
 ## Introduction
 
-This document describes the syntax and structure of the namespace for StarkNet..
+This document describes the syntax and structure of the namespace for StarkNet.
 
 ## Syntax
 
 The namespace "starknet" refers to the StarkNet Layer 2 chain on Ethereum.
+
+[Cairo field elements][] are one of the more novel aspects of Cairo's [Type system][], although future interfaces may not require any special knowledge or handling of field elements. ([More on field elements][].)
 
 ## References
 
--- a/starknet/README.md
+++ b/starknet/README.md
@@ -1,0 +1,31 @@
+---
+namespace-identifier: starknet
+title: StarkNet Ecosystem
+author: Argent Labs
+discussions-to:
+status: Draft
+type: Informational
+created: 2022-12-04
+updated: 2022-12-04
+replaces: CAIP-28
+---
+
+# Namespace for StarkNet
+
+## Introduction
+
+This document describes the syntax and structure of the namespace for StarkNet..
+
+## Syntax
+
+The namespace "starknet" refers to the StarkNet Layer 2 chain on Ethereum.
+
+## References
+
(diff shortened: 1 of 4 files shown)


bumblefudge:
More generally, you might think of CAIP10s and CAIP19s as "export formats" for addresses and assets in the system-- if there is a standardized and deterministic way to represent modulo-P/`felt` addresses and assets in modulo-2^256/"EVM-mode", that might be preferable to a lossful or arbitrary transformation of canonic representations

starknet/caip-2.md
```diff
@@ -0,0 +1,73 @@
...
+
+*For context, see the [CAIP-2][] specification.*
+
+### Rationale / Syntax
+
+The namespace `starknet` refers to the StarkNet Layer 2 on Ethereum.
+
+StarkNet currently represents its chains with string identifiers such as 
+`SN_MAIN` and `SN_GOERLI`. They also have field element representations which
+are not worth discussing in this specification.
```
bumblefudge, line 24:
```suggestion
`SN_MAIN` and `SN_GOERLI`. Internally, these strings are expressed as `field element` integers (according to a transformation documented in the system's [core documentation][Transaction structure]), but using the case-sensitive root string identifiers simplifies interoperability by sidestepping type system interoperability around integers.
```

starknet/caip-2.md
```diff
@@ -0,0 +1,72 @@
...
+# CAIP-2
+
+*For context, see the [CAIP-2][] specification.*
+
+### Rationale / Syntax
+
+The namespace `starknet` refers to the StarkNet Layer 2 on Ethereum.
+
+StarkNet currently represents its chains with string identifiers such as 
+`SN_MAIN` and `SN_GOERLI`. Internally, these strings are expressed as `field element` integers (according to a transformation documented in the system's [core documentation][Transaction structure]), but using the case-sensitive root string identifiers simplifies interoperability by sidestepping type system interoperability around integers.
```
iliav-starkware, line 23:
```suggestion
`SN_MAIN` and `SN_GOERLI`. Internally, these strings are encoded as field elements (according to a transformation documented in the system's [core documentation][Transaction structure]), but using the case-sensitive root string identifiers simplifies interoperability by sidestepping type system interoperability around integers.
```

delaaxe:
@bumblefudge I added a readme

starknet/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This document describes the syntax and structure of the namespace for StarkNet..
+
+## Syntax
+
+The namespace "starknet" refers to the StarkNet Layer 2 chain on Ethereum.
+
+## References
+
+[Type system]: https://starknet.io/docs/reference/syntax.html#type-system
```
ArielElp, line 25: Cairo types seem like a very low-level link in this context, and will also stop being relevant with Cairo 1.0, so I think that we can remove this.

PR #40, opened 2022-12-05
extends tezos/caip10 & caip122

QibingLee:
Adds CAIPs for Tezos.

### DIFF
--- a/tezos/caip10.md
+++ b/tezos/caip10.md
@@ -1,0 +1,55 @@
+---
+namespace-identifier: tezos-caip10
+title: Tezos Namespace - Addresses
+author: Qibing Li (@QibingLee)
+discussions-to: TBA
+status: Draft
+type: Standard
+created: 2022-12-06
+updated: 2022-12-06
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+_For context, see the [CAIP-10][] specification._
+
+## Rationale
+
+Tezos account is linked to a manager, which owns the public key. The hash of the public key outputs an address. Tezos supports the use of multiple public-key signature schemes, so the address prefix starts with `tz1` (Ed25519 curve), `tz2` (Secp256k1 curve), or `tz3` (P256 curve).
+
+## Syntax
+
+The syntax of a Tezos address matches the following regular expression:
+
+`(tz1|tz2|tz3) [1-9A-HJ-NP-Za-km-z]{33}`
+
+## Chain IDs
+
+_For context, see the [CAIP-2][] specification._
+
+| Network Name | Chain ID                         |
+| ------------ | -------------------------------- |
+| Mainnet      | NetXdQprcVkpaWU |
+| Devnet       | NetXm8tYqnMWky1 |
+
+
+## Test Cases
+
+```
+# Tezos Mainnet
+tezos:NetXdQprcVkpaWU:tz1MVqWiyfMbSESTaGEPxFjDWdM9zSXy1hcg
+
+# Tezos DelphiNet (Current active testnet)
+tezos:NetXm8tYqnMWky1:tz3gN8NTLNLJg5KRsUU47NHNVHbdhcFXjjaB
+```
+
+## References
+
+[Tezos Smart Contract](https://opentezos.com/tezos-basics/smart-contracts#general-definition-of-a-tezos-smart-contract)
+[CAIP-2]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md
+[CAIP-10]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md
+
+## Rights
+
+Copyright and related rights waived via CC0.
--- a/tezos/caip122.md
+++ b/tezos/caip122.md
@@ -1,0 +1,74 @@
+---
+namespace-identifier: tezos-caip122
+title: Tezos Namespace - SIWx
+author: Qibing Li (@QibingLee)
+discussions-to: TBA
+status: Draft
+type: Standard
+created: 2022-12-06
+updated: 2022-12-06
+requires: ["CAIP-122", "CAIP-2", "CAIP-10"]
+---
+
+## CAIP-122
+
+For context, see the [CAIP-122](CAIP-122) specification.
+
+## Rationale
+
+Tezos supports three types of keys, `tz1` for Ed25519 keys, `tz2` for Secp256k1 keys and `tz3` for P256 keys. This specification provides the signing algorithm to use, the `type` of the signing algorithm to identify it, and a method for signature creation and verification as required by [CAIP-122](CAIP-122).
+
+## Specification
+
+### Signing Algorithm
+
+In Tezos, Ed25519 (`tz1`) is the most commonly used signing algorithm. Tezos uses the base58 prefixes `edsk` for the Ed25519 private keys and `edpk` for the public keys to encode the keys. 
+
+### Signature Type
+
+We propose using the signature type `tezos:ed25519` to refer to the chain and algorithm used uniquely.
+
+### Signature Creation
+
+The abstract data model must be converted into a string representation in an unambigious format, and then the string converted to a byte array to be signed over.
+
+We propose the following string format, inspired from [EIP-4361](EIP-4361).
+
+```
+${domain} wants you to sign in with your Tezos account:
+${address}
+
+${statement}
+
+URI: ${uri}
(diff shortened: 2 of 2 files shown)


tezos/caip10.md
```diff
@@ -0,0 +1,55 @@
...
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+_For context, see the [CAIP-10][] specification._
+
+## Rationale
+
+Tezos account is linked to a manager, which owns the public key. The hash of the public key outputs an address. Tezos supports the use of multiple public-key signature schemes, so the address prefix starts with `tz1` (Ed25519 curve), `tz2` (Secp256k1 curve), or `tz3` (P256 curve).
```
zachferland, line 19: is there a similar distinction (to eth/eip155) in tezos between EOA (externally owned accounts) and contract based accounts? 

if contract based account here, can the manager change?

may want to describe each separately, and be able to identity both contract based account and eoa/keypair

tezos/caip10.md
```diff
@@ -0,0 +1,55 @@
...
+status: Draft
+type: Standard
+created: 2022-12-06
+updated: 2022-12-06
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+_For context, see the [CAIP-10][] specification._
```
zachferland, line 15: link/formatting

tezos/caip10.md
```diff
@@ -0,0 +1,55 @@
...
+tezos:NetXdQprcVkpaWU:tz1MVqWiyfMbSESTaGEPxFjDWdM9zSXy1hcg
+
+# Tezos DelphiNet (Current active testnet)
+tezos:NetXm8tYqnMWky1:tz3gN8NTLNLJg5KRsUU47NHNVHbdhcFXjjaB
+```
+
+## References
+
+[Tezos Smart Contract](https://opentezos.com/tezos-basics/smart-contracts#general-definition-of-a-tezos-smart-contract)
+[CAIP-2]: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md
```
zachferland, line 50: link formatting

tezos/caip122.md
```diff
@@ -0,0 +1,74 @@
...
+author: Qibing Li (@QibingLee)
+discussions-to: TBA
+status: Draft
+type: Standard
+created: 2022-12-06
+updated: 2022-12-06
+requires: ["CAIP-122", "CAIP-2", "CAIP-10"]
+---
+
+## CAIP-122
```
zachferland, line 13: for Ed25519, this caip-122 for this namespace looks good, since same message format and pretty standard Ed25519 stuff, but do you want to define it for Secp256k1 and P256 as well?

QibingLee: Actually, we can also define the Secp256k1 and P256, so did:tz1, did:tz2, or did:tz3 can be used. Then generated signature and its public_key should be grouped together, to ensure verification without any external APIs.

tezos/caip10.md
QibingLee: The manager is not the contract-based account. Actually, everything starts with a random number, then users can select different curves to generate different <public-key, private-key> pairs (tz1, tz2, tz3). They are totally different accounts to manage different funds. One advantage is, holding the random number makes you to take back funds across there accounts.

tezos/caip122.md
QibingLee: I have committed new changes to the Tezos namespace. Please check it for us

tezos/caip122.md
```diff
@@ -6,33 +6,33 @@ discussions-to: TBA
...
 
-Tezos supports three types of keys, `tz1` for Ed25519 keys, `tz2` for Secp256k1 keys and `tz3` for P256 keys. This specification provides the signing algorithm to use, the `type` of the signing algorithm to identify it, and a method for signature creation and verification as required by [CAIP-122](CAIP-122).
+Tezos supports three types of keys, `tz1` for Ed25519 keys, `tz2` for Secp256k1 keys and `tz3` for P256 keys. This specification provides the signing algorithm to use, the `type` of the signing algorithm to identify it, and a method for signature creation and verification as required by [CAIP-122][].
 
 ## Specification
 
 ### Signing Algorithm
 
-In Tezos, Ed25519 (`tz1`) is the most commonly used signing algorithm. Tezos uses the base58 prefixes `edsk` for the Ed25519 private keys and `edpk` for the public keys to encode the keys. 
+In Tezos, Ed25519 (`tz1`) is the most commonly used signing algorithm. Tezos uses the base58 with differnt prefixes to encode the public key, private key and signature. For Ed25519, the prefixes are `edpk`, `edsk` and `edsig`. For Secp256k1, the prefixes are `sppk`, `spsk` and `spsig`. For P256, the prefixes are `p2pk`, `p2sk` and `p2sig`.
```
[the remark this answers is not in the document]
QibingLee: Just public key, private key and signature. I just mentioned `tz1` as an example here.

bumblefudge:
Hey there @QibingLee -- For the most part this looks good, but I had a few questions:
1. Could we update CAIP-2 (and the chainIDs in CAIP-10 while we're here) with a Limanet and a Ghostnet example? If you could add, as I did in the CAIP-2 the genesis block hash, it helps people test their own computation of the chainID string for private chains or future testnets...
2. I'm a little lost on the mention of a public key being passed-- do dapps request (or already get from the wallet connection step) a public key, for both onchain and offchain use-cases? Should this public key be part of the SIWT message?
3. As above, could I ask you to provide a complete sample plain text message and a base64url-encoded byte representation of it, the way the [original CAIP-122 spec](https://chainagnostic.org/CAIPs/caip-122) does? This would serve the same "check your math" function as the CAIP-2 example in point # 1 above

bumblefudge:
Any updates, @QibingLee ? @zachferland would you feel qualified to address my requests instead, if Qibing isn't available to get this over the line?

tezos/caip10.md
```diff
@@ -0,0 +1,62 @@
...
+tezos:NetXm8tYqnMWky1:tz3gN8NTLNLJg5KRsUU47NHNVHbdhcFXjjaB
+```
+
+## References
+
+- [Tezos Smart Contract](https://opentezos.com/tezos-basics/smart-contracts#general-definition-of-a-tezos-smart-contract): General definition of a Tezos smart contract
+- [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md): Blockchain ID Specification
+- [CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md): Account ID Specification
+
+[CAIP-10]: https://github.com/ChainAgnostic/CAIPs/blob/8fdb5bfd1bdf15c9daf8aacfbcc423533764dfe9/CAIPs/caip-10.md
```
bumblefudge, line 56:
```suggestion
[CAIP-10]: https://ChainAgnostic.org/CAIPs/caip-10
```

jdsika:
Keep this open as I will likely have a look at it soon-ish

ukstv:
Hey all, re: CAIP-122, could you please provide an example of how a message should look like when presented to a user for signing? It would be interesting to see if there is any meaningful difference between tz1, tz2, tz3 signers.

bumblefudge:
@QibingLee  @jdsika  any update here?  I actually had a follow-up question that came up in another conversation about profiling CAIP-122:

1. what exactly are the valid possible values for `t` in the signature segment when serialized as a tripartite JWS object?

I see [here](https://github.com/ChainAgnostic/namespaces/pull/40/files#diff-5408e8364ec193f126ee987232fb8ef4a4aa5617f01804b44e039e1dd56e6553R28-R32) in the current draft that different binary objects get prefixed by type, but what I don't understand is the `signatureMeta.t` value. For comparison, the EVM equivalents are named after EIPs, `eip155` (for the classic `personal_sign` function) and `eip1271` (for smart-contract-wallet interfaces defined in that EIP), and they refer to interfaces that require different variables be passed.  The equivalent in the Tezos developer environment might be `taquitoInMemorySigner` and `beaconSigner`, since the digest being signed over has to be formatted/encoded differently for the different signing interfaces.  Ideally a CAIP-122 profile would define an ENUM of the possible signing-styles, so that when CACAOs get replayed and verified in a new context by a new verifier, they know how to encode the CAIP-122 message to make the signature verify 😄

bumblefudge:
@jdsika feel free to re-open if you are still working on Sign In With Tezos!

Magkoooh (approved):

PR #47, opened 2023-01-10
to address solana/caip2 inconsistencies pointed out online

bumblefudge:
Thanks to pythonmetaclass#4871 for pointing out the inconsistency between the namespace and the current Solana documentation/RPC behavior, and to AaronDGoldman#8819 and @ligi for helping troubleshoot the edits!

Pinging @oJshua & @antoineherzog for review.

### DIFF
--- a/solana/caip2.md
+++ b/solana/caip2.md
@@ -21,8 +21,8 @@ In CAIP-2 a general blockchain identification scheme is defined. This is the
 implementation of CAIP-2 for Solana. Blockchains in the "solana" namespace are
 validated by their genesis hash; each blockchain is maintained by a
 "[Cluster][]" of nodes, with its own gossip entrypoint and other endpoints.
-Since these genesis hashes are already 32-byte hashes, they require no
-transformations to be used as conformant CAIP-2 references.
+These genesis hashes require no transformations to be used as conformant CAIP-2
+references aside from concatenating their first 32 characters.
 
 ## Syntax
 
@@ -30,7 +30,7 @@ The namespace "solana" refers to the Solana open-source blockchain platform.
 
 ### Reference Definition
 
-The definition for this namespace will use the `genesisHash` as an indentifier
+The definition for this namespace will use the `genesisHash` as an identifier
 for different Solana chains. The method for calculating the chain ID is as
 follows with pseudo-code:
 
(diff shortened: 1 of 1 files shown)


solana/caip2.md
```diff
@@ -54,13 +56,13 @@ To resolve a blockchain reference for the Solana namespace, make a JSON-RPC requ
...
-  "result": "4sGjMW1sUnHzSxGspuhpqLDx6wiyjNtZAMdL4VZHirAn"
+  "result": "5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d"
 }
 ```
 
-The response will return as a value for the result a hash for the block with
-height 0 that should be truncated to its first 32 characters to be CAIP-30
-compatible.
+The response will return as a value for the result a 44-character
+Base-58-encoded hash for the block with height 0 that should be truncated to its
```
AaronGoldman, line 64: Do we care to distinguish base58btc from base58flickr?

https://github.com/multiformats/multibase/blob/master/multibase.csv

| encoding | code | description | status |
| -- | -- | -- | -- |
| base58btc | z | base58 bitcoin | default |
| base58flickr | Z | base58 flicker | candidate |

| base | alphabet |
| -- | -- |
| base58btc | `"123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"` |
| base58flickr | `"123456789abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ"` |

I believe it base58btc but didn't look it up.

bumblefudge: That is a very good question!
I am deducing that they use base58btc from the fact that [the crate they depend on](https://crates.io/crates/solana-program/1.8.6/dependencies) for base58 en/decoding is `[bs58](https://crates.io/crates/bs58/0.3.1)`, which allows multiple encoding alphabets but defaults to "the BITCOIN alphabet" ([src](https://docs.rs/bs58/0.3.1/bs58/alphabet/constant.DEFAULT.html)).  Will update the text and references now, since this could be determinant for implementers.

bumblefudge: this is also interesting, on the subject of base58btc and checksums:
https://github.com/solana-labs/solana/issues/6970

bumblefudge: resolving convo cuz PR updated

PR #50, opened 2023-01-19
Add caip2 for edeXa

ranjith-edx:
This PR adds a CAIP-2 definition for edeXa blockchain.

Reference of ChainAgnostic/CAIPs
#https://github.com/ChainAgnostic/CAIPs/pull/203

### DIFF
--- a/edeXa/README.md
+++ b/edeXa/README.md
@@ -1,0 +1,32 @@
+---
+namespace-identifier: edeXa
+title: Blockchain Reference for the edeXa Namespace
+author: Ranjith Kumar (@ranjith-edx), Shubham Koli (@shubham-edx)
+discussions-to: []
+status: Draft
+type: Informational
+created: 2023-01-19
+requires: 2
+---
+
+# Namespace for edeXa Blockchains
+
+This document defines the applicability of CAIP schemes to the blockchains of
+the edeXa business blockchain.edeXa Universe offers all well-known public blockchains in combination with our strong business blockchain technology to generate a new dimension of solutions
+
+## Syntax
+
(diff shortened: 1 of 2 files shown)


ligi:
Why don't you just use the eip155 namespace?

ranjith-edx:
@ligi  can you please elaborate  what your trying to say ...!?  so instead of edexa:1995 do you want me to add as the eip155:1995 itself..?!

ligi:
I ask why you do need a separate namespace and cannot just use eip155 - so what prevents you from just using eip155:1995

ligi:
OK - seems you need no separate namespace

bumblefudge:
@ranjith-edx -- I wrote this above but let me try being more explicit in case it helps.

if the ONLY way to connect to edexa is through the EVM JSON-RPC using chainID 1995, edexa is not a separate namespace, it is just a CAIP-2 value for the `eip155` namespace, i.e. the namespace of the EVM JSON-RPC and account system.  Namespaces are meant to describe entire namespaces, not branches or "layer 2s" within an existing set of languages and node interfaces.

Without doing further research, I'm inclined to agree with my co-editor that this should NOT be a separate namespace, as it seems to fall entirely in the eip155 namespace.  If the nodes are only accessible via EVM-compatible JSON-RPCs but there are, in fact, other idiosyncracies to the edeXa chain, such as, e.g., how CAIP-10 or CAIP-19 apply on chainID 1995, sections could be added to the eip155/caip10 and eip155/caip19 profiles.  Ping me on the CASA Discord if you need any further help documenting edeXa for EVM and non-EVM developers using the namespaces repo.

PR #52, opened 2023-01-24
Everscale

lamarketing:


everscale/README.md
```diff
@@ -0,0 +1,55 @@
+---
+namespace-identifier: everscale
+title: Everscale Ecosystem
+author: Lev Antropov(@levantropov), Vitaly Gritsay(@vvismaster)
+discussions-to: https://github.com/ChainAgnostic/CAIPs/pull/60
```
bumblefudge, line 5:
```suggestion
discussions-to: https://github.com/ChainAgnostic/namespaces/pull/52/
```

everscale/caip2.md
```diff
@@ -0,0 +1,82 @@
+---
+namespace-identifier: everscale-caip2
+title: EVERSCALE namespace
+author: Lev Antropov(@levantropov), Vitaly Gritsay(@vvismaster)
+discussions-to:
```
bumblefudge, line 5:
```suggestion
discussions-to: https://github.com/ChainAgnostic/namespaces/pull/52/
```

everscale/caip2.md
```diff
@@ -0,0 +1,82 @@
...
+
+In general, a transaction is generated with one incoming message (external or internal) and can generate several outcoming messages (external or internal) as a result.
+
+The transaction can be successful or aborted.
+
+For example, a simple value transfer consists of 2 transactions
+— Sender's transaction which generated
+an internal message with a value transfer,
+and Recipient's transaction where it received
+the message with value and updated its balance.
```
bumblefudge, line 53: I'm sorry, I'm not sure this section is relevant to CAIP-2 (it might be to a future CAIP we're working on!).  This section is supposed to tell a non-everscale wallet how to confirm that they are communicated to the correct endpoints for "mainnet" or "devnet".  You mention above that CAIP-2 identifiers map to the URLs of specific endpoints, so I'm assuming from the test case below that "mainnet" maps to `https://dapp01.itgold.io/graphql`. -- but **where** is that map authoritatively hosted? Is there a machine-readable registry somewhere that can be queried when a dapp tells a wallet to connect to `everscale:unfamiliarsubnet`? Should they by default not trust any CAIP-2 not listed in this profile, making THIS the authoritative registry/mapping?

The ### Resolution Mechanics section should tell the developer how to query `https://dapp01.itgold.io/` for confirmation that it is, in fact, `mainnet`-- many other RPC specifications include an `info` or `help` command which returns this identifier, while many that don't encourage wallets to query for the hash of block 0 or block 1 and use that as confirmation.  Please update this section to include that kind of guidance.

bumblefudge (changes requested):

everscale/caip2.md
lamarketing: Thanks, we update this part.

bumblefudge (approved):

PR #54, opened 2023-02-02
Fix Waves caip-10 test case for custom chain id

msmolyakov:
I just noticed that the address is derived not from the specified Chain ID.


bumblefudge:
hahahaha I noticed that too but forgot to request changes 🤦

bumblefudge (approved):

PR #62, opened 2023-03-17
Polkadot namespace CAIP-2 and CAIP-10 definitions

ntn-x2:
I used a rule we also use internally for better git integration: one-sentence-one-line.

I also added on what @bumblefudge already wrote, and clarified a few points that I did not find super clear. Glad to integrate any more feedback on the subject.

### DIFF
--- a/polkadot/README.md
+++ b/polkadot/README.md
@@ -1,11 +1,11 @@
 ---
 namespace-identifier: polkadot
 title: Polkadot Namespace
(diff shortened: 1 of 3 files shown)


bumblefudge:
This is excellent work-- merging and applying a commit straight to maini with minor editorial (such as the CAIP-2 prefixes for the CAIP-10 examples!)

PR #61, opened 2023-03-07
chore: fixes and improvements to solana caip10 and caip2

RyanTimesTen:
related to https://github.com/ChainAgnostic/namespaces/issues/60#issuecomment-1458440182

Issue #60, opened 2023-03-07
Potential Incorrect Information in Solana CAIP-10


RyanTimesTen:
Hello 👋 

I'm bringing up what I believe is a discrepancy between [Solana CAIP-2](https://github.com/ChainAgnostic/namespaces/blob/main/solana/caip2.md) and [Solana CAIP-10](https://github.com/ChainAgnostic/namespaces/blob/main/solana/caip10.md). Specifically, the Chain IDs do not match up, and I believe CAIP-2 has the correct values and CAIP-10 does not.

According to the [Chain ID Resolution Method](https://github.com/ChainAgnostic/namespaces/blob/main/solana/caip2.md#resolution-method) in CAIP-2:
> To resolve a blockchain reference for the Solana namespace, make a JSON-RPC request to the [RPC Endpoints](https://docs.solana.com/cluster/rpc-endpoints) of a blockchain node with method getGenesisHash, for example:
// Request
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "getGenesisHash"
}
// Response
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": "5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d"
}

And in fact, executing this resolution method on the Solana Mainnet RPC endpoint:
```
curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getGenesisHash"
}
'
```

yields this result
```
{"jsonrpc":"2.0","result":"5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d","id":1}
```

So the Chain ID for Solana Mainnet is `5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d`, as stated in CAIP-2, and the Chain ID stated in CAIP-10 is wrong. This is reproducible with the stated Chain ID for Solana Devnet.

Please let me know if I'm missing something, otherwise I'd be happy to open a Pull Request to correct this. Thanks!

bumblefudge:
Whoooops you are correct ser.  I think the CAIP-10 examples were using the [blockHash of block 0](https://explorer.solana.com/block/0) rather than the genesis hash proper, i.e.

```bash
curl https://api.mainnet-beta.solana.com -X POST -H "Content-Type: application/json" -d '
  {
[...]

PR #69, opened 2023-05-22
fix incorrect hash value in solana/caip19

lguild-plaid:
As addressed in https://github.com/ChainAgnostic/namespaces/pull/61 and referenced in https://github.com/ChainAgnostic/namespaces/issues/60#issuecomment-1458440182, some existing documentation was referencing an incorrect hash. While fixes were made for CAIPs 2 and 10, no fix was made for 19.

This PR updates the mainnet genesis hash used in solana CAIP-19 to the correct value.


bumblefudge (approved):
Good catch!

PR #72, opened 2023-06-06
add CAIP-10 for Algorand

k13n:
Add CAIP-10 definition for Algorand

### DIFF
--- a/algorand/caip10.md
+++ b/algorand/caip10.md
@@ -15,7 +15,7 @@ requires: ["CAIP-2", "CAIP-10"]
 
 ## Rationale
 
-Algorand addresses are canonically represented as 58-character long strings, written in all-upper-case. They are independent of the network in which they are used. Addresses are derived from a base32-encoding of a 32-byte key followed by a 4-byte checksum. The base32 padding (`=`) is dropped.
+Algorand addresses are canonically represented as 58-character long strings, written in all-upper-case. Account addresses are independent of the network in which they are used, and should be prefixed by the chainID of mainnet when stored in CAIP-10 format unless another chainID is significant in context (i.e. in a chainID-specific transaction). Addresses are derived from a base32-encoding of a 32-byte key followed by a 4-byte checksum. The base32 padding (`=`) is dropped.
 
 Algorand accounts, smart contracts, and logic signatures all share the same address format. They differ in how the 32-byte key is computed:
 
--- a/algorand/caip10.md
+++ b/algorand/caip10.md
@@ -1,0 +1,107 @@
+---
+namespace-identifier: algorand-caip10
+title: Algorand Namespace - Addresses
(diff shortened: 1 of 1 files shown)


ehanoc:
@bumblefudge @ligi Could we get :eyes: on this? Thanks

bumblefudge:
Apologies, both ligi and I were on leave the last few weeks.  Just looking at this now, I find the prose sections very clear and explanatory; my only concern is that the `chainId` segment is not optional in CAIP-10, and we do not have a wildcard (i.e. "equally valid on any chainID"), so I would recommend expressing all chainIDs as `mainnet` by default, with other values remaining valid in case you need them later.  My request changes should be enough, if you find them unobjectionable, but also feel free to rework/reword as you see fit and I'll review again.

algorand/caip10.md
```diff
@@ -0,0 +1,107 @@
...
+
+A raw address, consisting of the 32-byte key plus 4-byte checksum, contains `(32+4)*8 = 288` bits of information. The base32-encoding of this address contains 58 characters, each encoding 5 bits of information, hence `58 * 5 = 290` bits in total. This means the last two bits of the base32-encoding are redundant and Algorand enforces that they be zero. This means a valid Algorand address must end in one of the following characters: `[A, E, I, M, Q, U, Y, 4]`.
+
+
+## Syntax
+
+Algorand addresses match the following regular expression:
+
+```
+[A-Z2-7]{57}[AEIMQUY4]
```
bumblefudge, line 36:
```suggestion
algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73k:[A-Z2-7]{57}[AEIMQUY4]
```

algorand/caip10.md
```diff
@@ -0,0 +1,107 @@
...
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+*For context, see the [CAIP-10][] specification.*
+
+## Rationale
+
+Algorand addresses are canonically represented as 58-character long strings, written in all-upper-case. They are independent of the network in which they are used. Addresses are derived from a base32-encoding of a 32-byte key followed by a 4-byte checksum. The base32 padding (`=`) is dropped.
```
bumblefudge, line 18:
```suggestion
Algorand addresses are canonically represented as 58-character long strings, written in all-upper-case. Account addresses are independent of the network in which they are used, and should be prefixed by the chainID of mainnet when stored in CAIP-10 format unless another chainID is significant in context (i.e. in a chainID-specific transaction). Addresses are derived from a base32-encoding of a 32-byte key followed by a 4-byte checksum. The base32 padding (`=`) is dropped.
```

bumblefudge (changes requested):
Looks good, other than suggested caveat!

k13n:
Thank you, the suggested changes look good to me!

PR #75, opened 2023-06-30
Add Reef chain

boo-0x:


bumblefudge:
Hey @boo-0x thanks for the PR.  My main concern here is that I'm not clear on where the polkadot namespace ends and the reef namespace begins-- it feels more like Reef is an L1 using Substrate and the XCM that sits entirely in the Polkadot namespace, shares the same RPC standards and key formats and low-level engineering assumptions.  Then again, I have zero direct experience in PolkaDot so I'm probably oversimplifying here.  

And bear in mind, I'm *not* implying this is a trivial conceptual question or one that's obvious from a reading of [CAIP-104](https://chainagnostic.org/CAIPs/caip-104#what-is-a-namespace-reference). This is like the 5th time it's come up that a namespace contributor misses the nuance that multiple L1s can share a namespace.  Any help updating CAIP-104 to explain this better or give practical guidance on how to add additional L1s, Substrate pallets, etc to a shared base like Polkadot or Cosmos would be super, super appreciated (and I'm sure all future polkadot projects will thank you in particular!)

matjazonline:
Hello @bumblefudge, I'm Matjaz from Reef chain. Thank you for your reply, let me bring up some historic context to clear things up. Yes, Reef chain is built on Substrate framework the same framework that Polkadot and its parachains are running. In the past there was an effort to be part of Polkadot ecosystem as a parachain. That's why you found online articles about it. This didn't happen and Reef was never integrated into Polkadot and its XCM. 

So Reef chain is a separate L1 standalone chain with it's own RPC, governance and consensus mechanism that's why it would be natural to have distinct namespace.
If you have any further questions please let us know.

bumblefudge:
Interesting!  I'm thankful for the corner-case that brings me to make more explicit (and refine) our mental model of where the boundary lies between one namespace with divergent members and two namespaces!

So the way we've been thinking of namespaces is, as the name would imply:
1. a common set of types for data, keys, actors, etc.  for instance, is every valid Reef private key a valid Polkadot-ecosystem private key?  Does every Reef wallet address or contract address pass standard Polkadot validator functions in common tooling?
2. a finite set of namespaces for actors (wallets, deployed contracts, chainIDs) where collision could be confusing.  for example, are Reef contracts on Reef chains addressable using XCM [v3]?
3. a "virtual machine" or runtime in common? Are L2 or variant-L1 nodes running the same code with extenions, or different code?  (The Rusty crate-based composability of Substrate already stretches this definition a bit, which is very interesting relative to the Ethereum's world's battle for a precise definition of "EVM compatibility"!)
4. a common set of interfaces for key components like Wallets (Ethereum isn't the best example of a standardized wallet interface just yet!), Nodes (is node functionality defined as a [Swagger/OAS3](https://support.smartbear.com/swaggerhub/docs/tutorials/openapi-3-tutorial.html) or an [OpenRPC](https://open-rpc.org/) spec if it's JSON-RPC based?) Is Reef's a superset of Polkadot's, or are there incompatibilities like common functions behaving differently?

I hope this doesn't come across as an inquisition, I just find it hard to trust my own judgment judging the above questions from my own google-research and would rather have a Reef engineer confirm my hunches on the above!

bumblefudge:
(And in case you can't tell, the above will be hardened into a PR on CAIP-104 and an issue template on this repo once we've dogfooded this intake interview methodology :D )

matjazonline:
Hey @bumblefudge, thanks for quick response and interesting for me as well to see all the criteria for a namespace. Substrate is just a framework and very modular so chains built with it can be very different technically. Where Polkadot is an implemented network of chains (parachains) that share block validation and XCM. 
So to answer the above:

 1) types for data are different. Since we use the same type of cryptography private key to generate public address can be shared. But it generates Reef specific SS58 formatted public address and can not be used on Polkadot chain. So in this way accounts are different and each Substrate chain can choose different SS58 format but still validation tools can be the same provided correct SS58 format.
 2) Official Reef wallet is standalone designed to work with Reef chain only. Deployed contracts are not addressible via XCM. XCM is only available between connected parachains ( like Polkadot ). Reef is a standalone L1 and we need cross chain bridge to connect to other chains. Also the REEF coin has 18 decimals while DOT has 10.
 3) Runtime is a collection of Substrate modules and is unique to Reef chain. We have modified or added our custom implementation to the runtime. With Substrate upgradability we can additionally upgrade/modify runtime without forking the chain in the future.
 4) The Substrate modules collectively form a RPC interface so they are unique based on implementation. RPC is not strictly JSON-RPC out-of-the-box since responses are not in JSON but encoded with SCALE codec. Every Substrate chain has a unique metadata description that describes endpoints and its types and a unique types definition so they can be decoded. So Reef RPC is not a superset of Polkadot - some methods are unique to Reef, some might have same name but they have different parameters and vary in implementation or types. Reef does not accept all Polkadot RPC calls and you can not expect to have compatible RPC calls between Reef or Polkadot.
 
 As times go on there's many more options to build a blockchain and you have harder time to comprehend them to make a solid decision :) Hope this was helpful, thanks!

bumblefudge:
Wow, that's super cool! I didn't realize you guys had undertaken such a big lift making a totally distinct namespace! Very cool.  I would recommend including the details of this SS58 stuff when you write a CAIP-10 profile, and including a link to a machine-readable RPC dictionary (say, an OpenAPI3 or OpenRPC document) when you write a CAIP-27 profile.

I added [a commit](https://github.com/ChainAgnostic/namespaces/pull/75/commits/3200f987c940f60730b329ed7f0299f31104f289) with some editorial suggestions, feel free to keep pushing more commits and re-requesting review if you disagree with any of them or want to keep adding stuff.  I know at first glance it might seem counter-intuitive to mention Polkadot in the readme, but trust me, people who don't know the history and just see "Substrate" will assume all the same things I did unless they see your most recent response!

bumblefudge (approved):

boo-0x (commented):
Looks good to me

boo-0x (commented):
LGTM

matjazonline (approved):

PR #77, opened 2023-07-10
adding vechain caip-2

darrenvechain:


grenos (approved):
legit!

akanoce (approved):

Agilulfo1820 (approved):
LGTM

vechain/README.md
```diff
@@ -0,0 +1,29 @@
...
+This document defines the applicability of CAIP schemes to the blockchains of
+the VeChain ecosystem.
+
+## Syntax
+
+The namespace "vechain" refers to the VechainThor open-source blockchain platform.
+
+## References
+
+- [VechainThor Resources](https://www.vechain.org/vechainthor/)
```
bumblefudge, line 23:
```suggestion
- [VechainThor Resources](https://www.vechain.org/vechainthor/)
- [VeChain Developer Resources](https://docs.vechain.org/development-resources)
```

vechain/README.md
```diff
@@ -0,0 +1,29 @@
...
+- [VeChain Developer Portal](https://docs.vechain.org/)
```
bumblefudge, line 24:
```suggestion
- [VeChain Developer Portal](https://docs.vechain.org/)
- [VeChain Developer Portal Self-Test Resources](https://docs.vechain.org/development-resources)
```

bumblefudge:
Hey, this all looks great and factual and helpful to non-VeChain-experience devs.  It's not required for approval but a few things I *would* like to see added, if you have the bandwidth to drop in some links or a sentence or two of explanation, for the sake of developer education:
- CAIP-10 - What is the relationship between private keys and addresses? Are addresses just bare public keys, hashes of public keys, etc? If a hash, can the public key be recovered with from the signature of a known payload like secp/eth keypairs using ECRecover, or are there other ways to calculate it?
- CAIP-10 - Does the same private key produce the same address on both of today's chains (CAIP-2s) and all future VeChains? Or are addresses chain-specific?
- CAIP-2 - it might help to add a section explaining VORJ's EVM/Layer2 mode, linking to the [canonical CAIP-2 data within the EIP155 namespace](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-100009.json) and explaining the limitations of EVM-mode versus native mode (i.e., an arbitrary contract `0x1234deadbeef` on Ve mainnet can't be interacted with on `eip155:100009:0x1234deadbeef`, can it?).  If a subset of VeChain can be interacted with via eip155:100009, maybe explain what subset, and how a developer could confirm membership in it before sending data to it that might timeout, etc?

bumblefudge (approved):

PR #84, opened 2023-08-15
feat: Adds CAIP-122 standard for XRP Ledger.

antondalgren:
Aligned with the ripple-keypairs implementations for signing and verification.

### DIFF
--- a/xrpl/caip122.md
+++ b/xrpl/caip122.md
@@ -21,9 +21,10 @@ This specification provides the signing algorithm to use, the `type` of the sign
 
 ### Signing Algorithm
 
-XRPL uses the ECDSA [secp256k1][secp256k1] or  EdDSA [Ed25519][Ed25519] signing algorithm for signing and verifying messages depending on the type of the keypair that is being used. A [secp256k1][secp256k1] keypair will be using the [secp256k1][secp256k1] algorithm and a [Ed25519][Ed25519] keypair will be using the [Ed25519][Ed25519] algorithm.
+XRPL uses the ECDSA [secp256k1][] or  EdDSA [Ed25519][] signing algorithm for signing and verifying messages depending on the type of the keypair that is being used. A [secp256k1][] keypair will be using the [secp256k1][] algorithm and a [Ed25519][] keypair will be using the [Ed25519][] algorithm.
 
-Prior to the message being signed using [secp256k1][secp256k1], it needs to be hashed using the SHA-512/2 algorithm. Correspondingly it needs to be converted to bytes prior to signing it using the [Ed25519][Ed25519] algorithm.
+Prior to the message being signed using [secp256k1][], it needs to be hashed using the 512bit digest of the SHA-2 algorithm, commonly referred to as [SHA-512][] and specified in [RFC4634][]. 
+Correspondingly, a message needs to be serialized to bytes prior to signing it with the [Ed25519][] algorithm.
 
 ### Signature Type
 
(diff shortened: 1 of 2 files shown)


antondalgren:
Any news on this @bumblefudge?

bumblefudge:
LGTM with minor edits, which i left on [your fork](https://github.com/TowoLabs/namespaces/pull/3).  Are there already public and open source implementations of this that you can link to from the # References section?  If so, please add in the form:

[name of implementation]: link to git repository
 {...}
- [name of implementation]: Sample implementation by {organization(s)} [, live at {url}]

bumblefudge (approved):

xrpl/caip122.md
```diff
@@ -0,0 +1,112 @@
...
+
+The preceeding example would be hashed to `4beb3f7d5c8bf8deab467de033628274a9053f688f82e0f023cddcb0564cfc9a` prior to being signed using the [secp256k1][] algorithm.
+
+The serialization of this string prior to signing with [Ed25519][Ed25519] would produce:
+```
+736572766963652e6f72672077616e747320796f7520746f207369676e20696e207769746820796f7572205852504c206163636f756e743a0a72344654766e6168625566686531574b32454b354a7a34634e766446765438447a740a0a49206163636570742074686520536572766963654f7267205465726d73206f6620536572766963653a2068747470733a2f2f736572766963652e6f72672f746f730a0a5552493a2068747470733a2f2f736572766963652e6f72672f6c6f67696e0a56657273696f6e3a20310a436861696e2049443a20300a4e6f6e63653a2033323839313735370a4973737565642041743a20323032312d30392d33305431363a32353a32342e3030305a0a5265736f75726365733a0a2d20697066733a2f2f516d653773733341525667787636725871565069696b4d4a3875324e4c676d67737a673133705972444b456f69750a2d2068747470733a2f2f6578616d706c652e636f6d2f6d792d776562322d636c61696d2e6a736f6e
+```
+### Signature Verification
+
+Signature verification behaves similarly. We can use standard [secp256k1][] or[Ed25519][] signature verification.
```
bumblefudge, line 88: Note that in the case of EVM, there are multiple signing methods (the spec mentioned eip155, aka `personal_sign`, classic metamask arbitrary unstructured text signing, and eip1271, aka `smart contract wallet signing`, for signatures produced by deployed smart contracts).  is there only one signing method in Ripple, that you simply pass a secp256k1 or ed25519 key to, or are they two different signing interfaces with slightly different digest constraints/encodings, etc? 

if the latter, it would be good to define the ENUM of valid `signatureMeta.t` values (recognizable names of signature interfaces in the XRP toolchain). and if the former, maybe just an ENUM with only one member? 😄

bumblefudge: https://xrpl.org/signing-methods.html

I hadn't thought about multi-signatures, but maybe we just include that in the enum and let future teams worry about that one 😅

antondalgren: There is only one signing method in Ripple, that you simply pass a secp256k1 or ed25519 key.

xrpl/caip122.md
```diff
@@ -30,6 +30,16 @@ Correspondingly, a message needs to be serialized to bytes prior to signing it w
...
+
+```
+type SignatureMeta struct {
+  SigningPubKey String
+}
+```
+
 ### Signature Creation
 
 The abstract data model must be canonicalized to a string representation in an unambigious format, and then the string serialized to the proper binary format mentioned [above](#signing-algorithm) to be signed over with the proper algorithm.
```
bumblefudge, line 45: Would it help to just specify JCS here, since it's the most general-purpose canonicalization for JSON?

xrpl/caip122.md
```diff
@@ -30,6 +30,16 @@ Correspondingly, a message needs to be serialized to bytes prior to signing it w
 
 We propose using the signature type `xrpl:secp256k1` and `xrpl:ed25519` to refer to the chain and algorithm used uniquely.
```
bumblefudge, line 31: I would maybe add a sentence to this paragraph like, 
```
These two signature types are the only two available and can be considered the defaults per key type. The typing of the signature MAY be made explicit by including a `SignatureMeta.t` property set to either `xrp1:secp256k1` or `xrp1:ed25519`, but if omitted the appropriate one of these two defaults should be obvious from the keytype provided in the `SignatureMeta.SigningPubKey` value.
```

xrpl/caip122.md
bumblefudge: i mean, multi-signature might be the same algorithm/method but isn't the payload a little differently constructed? or am i missing something? you don't need to define it, I just want to make sure someone else defining it later understands how to do it (i.e., how to PR this adding a section for what's particular to it and adding a new valid value for `SignatureMeta.t`)

xrpl/caip122.md
antondalgren: I have no strong opinions regarding adding it or not.

xrpl/caip122.md
antondalgren: I think the context of your sentence is covered in the `Signing Algorithm` section.

xrpl/caip122.md
antondalgren: I don't think multi-signature is relevant for this use case. But I'm happy to learn more about it otherwise.

obstropolos (approved):

PR #97, opened 2023-10-19
[Errata arweave/caip122] - explicitly add signatureMeta.t value

bumblefudge:
Nota bene @ropats16 @joshbenaron @danmacdonald @zachferland :
Implementer feedback on CAIP-122 pointed out that to me that profiles need to explicitly NAME signature types so that they can be referenced in each CACAO conforming to that name. We made it [more explicit for future profilers](https://github.com/ChainAgnostic/CAIPs/pull/247/files#diff-c0928c7ab9eb00925c752323139aa16e42a74e5a3720bbab0bf7b4cb78ca0dbdR55-R59) but now we're also going back and inserting one for this CAIP-122 profile.  the name was chosen out of a hat so please "request changes" if you can think of a more precise, descriptive, and/or future-proof one!

### DIFF
--- a/arweave/caip122.md
+++ b/arweave/caip122.md
@@ -71,7 +71,7 @@ Type: ${type}
 Signature: ${signature}
 ```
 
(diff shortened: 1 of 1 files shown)


obstropolos (approved):

bumblefudge:
cant self-approve, override rules and merging

PR #101, opened 2024-02-06
initial structure for eip155/caip25 and /caip211 profiles

bumblefudge:
gotta have some basic precedents so that other namespaces can define THEIR caip-25 profiles in turn!

### DIFF
--- a/eip155/caip211.md
+++ b/eip155/caip211.md
@@ -1,0 +1,51 @@
+---
+namespace-identifier: eip155-caip25
+title: EIP155 Namespace, aka EVM Chains - JSON-RPC Provider Authorization
+author: Pedro Gomes (@pedrouid), Hassan Malik (@hmalik88), bumblefudge (@bumblefudge)
+discussions-to: 
+status: Draft
+type: Standard
+created: 2024-02-05
+updated: 2024-02-05
+requires: ["CAIP-25", "CAIP-211", "CAIP-217"]
+---
+
+# CAIP-25
+
+_For context, see the [CAIP-211] and [CAIP-25] specifications._
+
+## Rationale
+
+In the Ethereum space, JSON-RPC connections are not constrained by a single interface definition; numerous overlapping implicit and explicit authorities define the valid interactions between a user agent like a wallet and a decentralization application (whether web-based or otherwise).
+The core RPC capabilities (concretely, the syntax and semantics of RPC methods and notifications) of any Ethereum or EVM-compatible network are assumed to be defined by the consensus of the Ethereum Improvement Proposal process.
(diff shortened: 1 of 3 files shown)


eip155/caip25.md
```diff
@@ -0,0 +1,51 @@
...
+
+In the Ethereum space, many networks interact in a given application or session, and as such it is necessary for applications and user-agents like wallets to negotiation and authorize interfaces dynamically and mutually.
+The [CAIP-25] negotiation allows for multiple [CAIP-217] authorization scopes to be negotiated together. 
+
+## Network-specific versus Namespace-wide Scopes
+
+Authorizing permissions and capabilities to each network separately (i.e., in its own [CAIP-217] authorization object) is recommended in most use-cases, so that over the session and across sessions, each network can granularly gain or attenuate permissions and capabilities separately. 
+Conversely, an authorization object scoped to all of `eip155` can only add or substract individual networks from the `chains` array, or add and substract capabilities to all enumerated networks in that array.
+
+It is also worth mentioning that if a network supports any capabilities NOT supported by other networks, they should never share an authorization object, as the semantics of [CAIP-217] interpret this as supporting EVERY capability lited on EVERY network listed.
```
hmalik88, line 27:
```suggestion
It is also worth mentioning that if a network supports any capabilities NOT supported by other networks, they should never share an authorization object, as the semantics of [CAIP-217] interpret this as supporting EVERY capability listed on EVERY network listed.
```

eip155/caip211.md
hmalik88, line 1: Not sure if we need a separate file for this? I feel like this can be a subsection under the caip25 doc

bumblefudge: i'm just being pedantic because it's technically it's own separate CAIP... maybe I should link to it from the 25 one, though, I think you're right that it could get missed!

bumblefudge:
@hmalik88 does the new cross-link look ok? technically i need a 2nd not-me reviewer :D

eip155/caip25.md
```diff
@@ -0,0 +1,113 @@
...
+      },
+      "cosmos": {
+        ...
+      }
+    },      
+    "sessionProperties": {
+        "expiry": "2022-11-31T17:07:31+00:00"
+        "capabilities": {
+            "0x31": {
+                "paymasterService": {
```
arein, line 78: would this be a [`pm_getPaymasterData`](https://github.com/ethereum/ERCs/pull/360/files) or where would the values for capabilities be defined?

bumblefudge: Yes, but I didn't link to it because it's not merged yet!
https://github.com/ethereum/ERCs/pull/360/files

PR #102, opened 2024-02-07
solana/caip25.md initial draft

bumblefudge:
Only in draft because I'm waiting on real-world examples for the ## Examples section from a prototyping team. Big thanks to the many contributors so far.

Can't assign for review without adding you to the org but @jnwng @arein @glitch review from any of you would be great.

### DIFF
--- a/solana/caip25.md
+++ b/solana/caip25.md
@@ -1,7 +1,7 @@
 ---
 namespace-identifier: solana-caip25
 title: Solana Namespace - JSON-RPC Provider Authorization
-author: Jon Wong (@jnwng), Glitch (@glitch) bumblefudge (@bumblefudge)
+author: Jon Wong (@jnwng), Glitch (@glitch-txs) bumblefudge (@bumblefudge)
 discussions-to: https://github.com/anza-xyz/wallet-adapter/issues/807
 status: Draft
 type: Standard
--- a/solana/caip25.md
+++ b/solana/caip25.md
@@ -30,12 +30,12 @@ Multi-chain use-cases incorporating non-Solana chains should use distinct [CAIP-
 Wallet capabilities in the [Wallet Standard] model are declared upfront at time of connection and persisted as read-only constants, so wallets are recommended to declare them explicitly in the `sessionProperties` object in a [CAIP-25] interaction for maximum interoperability and stability.
 Note that the `sessionProperties` object is shared across any and all namespaces in a given wallet<>application session, which may span multiple types of networks all declaring `sessionProperties`, so to avoid collisions the following properties names are recommended for easy mapping to the **case-sensitive** properties names in the [Wallet Standard]:
 
-|`sessionProperties` key in [CAIP-25] expression|[WalletAdapter] mapping|values|documentation|reference implementation|
+|`sessionProperties` key in [CAIP-25] expression|[Wallet Adapter] mapping|values|documentation|reference implementation|
 |---|---|---|---|---|
 |`SolanaTransactionVersion`|`SolanaTransactionVersion`|`0` or `legacy` (default)|[solana.com][Transaction Version Declaration]|[github.com][Transaction Version Refimpl]|
 |`SolanaTransactionCommitment`|`SolanaTransactionCommitment`|`finalized` (default), `confirmed`, `processed`|[solana.com][State Commitment Declaration]|[github.com][State Commitment Refimpl]|
 
-As these properties are mandatory and required for stability in [Wallet Standard] connections, [CAIP-25] providers MAY inject defaults for all of the above based on prior knowledge or out-of-date registries.
+As these properties are mandatory and required for stability in [Wallet Standard] connections, [CAIP-25] providers MAY inject defaults if any of the above are not set explicitly by a wallet.
 For this reason, wallets are RECOMMENDED to set them explicitly even if not requested by their counterparty to avoid inaccurate values being injected.
 
 ## Examples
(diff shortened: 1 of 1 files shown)


solana/caip25.md
```diff
@@ -0,0 +1,69 @@
+---
+namespace-identifier: solana-caip25
+title: Solana Namespace - JSON-RPC Provider Authorization
+author: Jon Wong (@jnwng), Glitch (@glitch) bumblefudge (@bumblefudge)
```
glitch-txs, line 4:
```suggestion
author: Jon Wong (@jnwng), Glitch (@glitch-txs) bumblefudge (@bumblefudge)
```

124maa (changes requested):
[OK](url)

bumblefudge:
@0xproflupin anything jump out at you as patently wrong? @obstropolos said you might know a thing or two about Solana interop

glitch-txs (approved):

obstropolos (approved):
No major changes / looks good - just make sure before merging: 
- Add a comma on the author list 
- Watch for capitalization consistency in headers 
- Pick "multichain" vs "multi-chain" / consistency

vsakos:
@bumblefudge If I understand the spec correctly, the `SolanaTransactionVersion` and `SolanaTransactionCommitment` fields are locked to a specific value for the whole dApp <-> wallet session.

I don't think that declaring a transaction version makes sense for the whole session, because most wallets support both versioned (only v0 for now, but we will probably have v1, v2, etc. in the future) and non-versioned (legacy) transactions, and the dApp might want to use both versions in the same session. So, limiting the session to a single version might not be ideal. That's the reason why the Wallet Standard exposes the `supportedTransactionVersions` field, so the dApp knows which versions are supported by the wallet, but is free to use any.

As for the commitment, unlike other chains, dApps on Solana are expected to use their own JSON-RPC endpoints for fetching data. Currently, the only method that can accept the commitment is the `signAndSendTransaction` method. Just like with the transaction version, the dApp might want to use different finality for different transactions. For example, they can use `confirmed` for low-risk transactions to have transaction confirmations faster and `finalized` for high-risk transactions (like transferring larger token amounts) where it's acceptable to wait longer for the transaction to be confirmed.

bumblefudge:
> @bumblefudge If I understand the spec correctly, the `SolanaTransactionVersion` and `SolanaTransactionCommitment` fields are locked to a specific value for the whole dApp <-> wallet session.
> 
> I don't think that declaring a transaction version makes sense for the whole session, because most wallets support both versioned (only v0 for now, but we will probably have v1, v2, etc. in the future) and non-versioned (legacy) transactions, and the dApp might want to use both versions in the same session. So, limiting the session to a single version might not be ideal. That's the reason why the Wallet Standard exposes the `supportedTransactionVersions` field, so the dApp knows which versions are supported by the wallet, but is free to use any.

I might have made a terrible mistake, in that case-- the wallet should be declaring `supportedTransactionVersions` not `SolanaTransactionVersion` (must have misread the Solana specs here).  The intention is not to force the dapp to always use one or the other version, but to tell WC upfront what value to expose to the dapp via Wallet Standard.

> As for the commitment, unlike other chains, dApps on Solana are expected to use their own JSON-RPC endpoints for fetching data. Currently, the only method that can accept the commitment is the `signAndSendTransaction` method. Just like with the transaction version, the dApp might want to use different finality for different transactions. For example, they can use `confirmed` for low-risk transactions to have transaction confirmations faster and `finalized` for high-risk transactions (like transferring larger token amounts) where it's acceptable to wait longer for the transaction to be confirmed.

Same-- I thought the wallet might have some way of constraining or suggesting only `finalized`, for example, but I might also have misread the Solana docs here as well on what the wallet has the option to request or constrain.  Should this be dropped altogether from the things a wallet can declare, and is purely a dapp<>RPC concern?

vsakos:
@bumblefudge If the idea is to be fully compatible with the Wallet Standard, I think `SolanaTransactionVersion` (or some alternative name) should be a list of available versions, just like in the WS itself, instead of a single transaction version.

From the dApp's standpoint, a Solana wallet is just a signer and nothing more - with the exception of `signAndSendTransaction`, which is the only method that makes use of the commitment param. However, most of the Solana dApps still choose to use `signTransaction` and relay the transaction to their own RPC instead of relying on the wallet to do it. But in any case, the practice is to fully honor the dApp's request (with default values and fallbacks for missing params etc.)

glitch-txs:
Thanks so much @vsakos for bringing this up! Let's revisit this, could we open a new PR for modifications? @bumblefudge 

We haven't started implementing this in the WalletConnect protocol yet, but it would be nice to start ASAP once this is ready to go to prod 🙏

bumblefudge:
> @bumblefudge If the idea is to be fully compatible with the Wallet Standard, I think `SolanaTransactionVersion` (or some alternative name) should be a list of available versions, just like in the WS itself, instead of a single transaction version.

you're right, it seems I completely missed that the wallet exposes [an array](https://github.com/wallet-standard/wallet-standard/blob/ba5c16b7f608754ff5c494079f6b7549b857c390/packages/example/wallets/src/solanaWallet.ts#L81), not a maximum-value (with the assumption of all lower/older values being implicit).

what I find odd is that in the example linked above, where, as you say, commitment _would in fact matter_ and a dapp might want to know what commitments it could send via the wallet, the example doesn't include that variable... is that declared elsewhere? or does the wallet standard not contemplate wallets supporting a subset of commitment modes in the first place?

vsakos:
@bumblefudge The commitment definitely matters, but I would not limit it in any way. Wallets should support everything the blockchain itself supports. After all, it's just an enum/string that gets forwarded to the RPC. https://solana.com/docs/rpc#configuring-state-commitment.

The commitment itself is just an indicator of which state of the blockchain should be considered for reading data or relaying a transaction. For example, if you are receiving some tokens in transaction 1 and you want to send those tokens to another wallet with transaction 2, and if tx1 is confirmed but does not have enough confirmations to be considered finalized, if you relay tx2 with confirmed it will go through but it will fail with finalized commitment because you still don't have those tokens in the finalized state.

This is why, in my opinion, it makes more sense to make this parameter per request/operation and not limit the dApp or the wallet in any way.

bumblefudge:
@vsakos -- eloquently explained, thanks so much.  I came to more or less the same conclusion after re-reading the RPC docs and the wallet adapter repo.  I tagged you in the update PR, but it's pretty much what you recommended here and nothing new/extra beyond that.

PR #117, opened 2024-06-25
Add Alephium

h0ngcha0:


nop33:
Hey @bumblefudge, a long time ago you [commented](https://github.com/alephium/alephium-frontend/issues/158#issuecomment-1845735619) that we could reach out to you once we are ready! This is a PR to integrate Alephium. Let us know if you have some time to check it out! Thanks a ton!

alephium/caip10.md
```diff
@@ -0,0 +1,74 @@
...
+- **P2C** - serialized contract id
+
+Hash algorithm used is blake2b.
+
+Constructing an address:
+
+- **address** = `address type || content bytes`
+
+Each address on Alephium belongs to a group, which can be derived
+deterministically from the address.
```
bumblefudge, line 43: This is interesting (and reminds me of a similar property of Polkadot addresses) but for the CASA readership you need to spell this out a little more: link to (and summarize as concisely as possible) _how_ you derive group membership from the address (and whether an address is specific to mainnet/devnet or not, and whether that too can be derived diagnostically from an address).

alephium/caip10.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Syntax
+
+Alephium address is represented as a Base58 encoded string.
```
bumblefudge, line 47:
```suggestion
The native Alephium address encoding is a [Base58][]-encoded string of the type defined above.
[CAIP-10][] simply uses an entire native address as the `account_address` component.
```

bumblefudge: But this begs the question... what is the [chain_id segment](https://chainagnostic.org/CAIPs/caip-10#syntax) for each CAIP-10 example listed below in the test vectors?

You mention above that group membership can be derived from the address, but it was unclear how.  Do the address types correspond to group 0, 1, 2, and 3 respectively?

I was also unclear if each address "exists" only on a specific network (i.e., on mainnet or devnet), or if an address is equally valid on all 3, even if it only has history on 1 or 0 of them.

The CAIP-10 system is heirarchical and most CAIP-10 implementations have a usecase or two requiring them to know (if not a universal need to know) _which network to query for history_ of a given account/address, which is why there is a CAIP-2 embedded in each CAIP-10.
If this is not the case and addresses are universal, you might want to add a "-1" to the enum of networks you define in CAIP-2, like `offchain` or `universal`. This way, all CAIP-10s could be `alephium:universal/-1:{address}`.
Conversely, if there IS no universal/offchain address namespace in the app development model of alephium, do NOT introduce such a CAIP-2 wildcard but explain how a given address should be namespaced/mapped to CAIP-2s.

Fascinating stuff here, I love how different this is from other CAIP-2 and CAIP-10 profiles!

alephium/caip19.md
```diff
@@ -0,0 +1,46 @@
...
+bytes blake2b hash encoded as a hex string.
+
+## Syntax
+
+Alephium token id represented as 32 byte array encoded as hex string.
+
+## Examples
+
+```
+383bc735a4de6722af80546ec9eeb3cff508f2f68e97da19489ce69f3e703200
```
bumblefudge, line 35: Here too I'm confused about the missing `alephium:` prefix and CAIP-2 segment!

bumblefudge (changes requested):
This is looking great so far, but the CAIP-10 and CAIP-19 profiles aren't quite conformant, and I am a little too pressed for time to do enough research to recommend more concretely what next steps would make for the best possible CAIP-2 segment to add to them!

alephium/caip10.md
h0ngcha0: Thanks for the great comment!

> You mention above that group membership can be derived from the address, but it was unclear how. Do the address types correspond to group 0, 1, 2, and 3 respectively?

I added a description in this [commit](https://github.com/ChainAgnostic/namespaces/pull/117/commits/3afba3c2bfd832319e29de8af9950f59c81b16f0), basically the group information is encoded as part of the address and can be derived from it. 

Thanks for the context for `CAIP-2`. The address is indeed "universal" in the sense that it is valid for all networks, and the group information is implicitly embedded in the address. However dApps on Alephium can request an address from a certain group on a certain network. For example, a testnet dApp can request connection on chain id `alephium:testnet/1`, the wallet should connect the dApp with an address from group `1` on testnet. If a mainnet dApp request connection on chain `alephium:mainnet/-1`, then the wallet can connect the dApp with an address from any groups on mainnet.

So I am a bit unsure here as well, I specified chain id in`CAIP-2` as:
```
alephium + ":" + network_id + "/" + group
```
Because I thought dApp can use this format to request address from a certain group (or all groups) on a certain network, but since address is universal and group is already implicitly embedded in address, if feels like the address should be specified like `alephium:universal:{address}`. 

Really appreciate it if you could give us some advice here :pray:

alephium/caip19.md
h0ngcha0: Tokens are "universal" as well in the sense that they are not tied to a specified network. It should probably be something like `alephium:universal:{tokenId}`, but I am not sure about how to reconcile with the `CAIP-2` part in this case, similar to address.

h0ngcha0:
> This is looking great so far, but the CAIP-10 and CAIP-19 profiles aren't quite conformant, and I am a little too pressed for time to do enough research to recommend more concretely what next steps would make for the best possible CAIP-2 segment to add to them!

Thanks a lot for your time! 

Do you think the following would work?

1. For chain id, we use `alephium:${network_id}/${group}`, because Alephium is a sharded blockchain that contains a number of smaller chains internally, so we want to be able to specify the specific "chain" inside of it through group. 

2. For address, we specify `alephium:${network_id}:${address}`. Because `address` already have the group information implicitly we don't need to specify it. 

What do you think? Any advice would be appreciated, thanks!

alephium/caip19.md
bumblefudge: yeah, that makes sense! if something is shard-specific or chain-specific it would go in the `chainId` reference (the second segment), and if you want to create a special case for `universal` addresses, you just add it to the caip2 profile "specification" section!

alephium/caip2.md
```diff
@@ -0,0 +1,61 @@
...
+into distinct groups. The blockchain id includes both the network id
+and the group information.
+
+The Network id can be one of three values: `mainnet`, `testnet` and
+`devnet`.
+
+Currently, Alephium has 4 groups. The group component of the
+blockchain id can take one of five values: `0`, `1`, `2` and `3` which
+correspond to groups `0` through `3`, and `-1` which represents any
+group.
```
bumblefudge, line 29: I think what's missing here is an explanation of what a group _does_, and in what context "any group" would be meaningful.  for example: is there a 1:1 relationship between blockchains and groups? Does each group contain 4 blockchains, and each blockchain belong only to one group? 

It might also be worth mentioning _somewhere_, maybe in CAIP-19 or CAIP-10 or both, that smart-contracts are group-specific and that most wallets control an address in each group for this purpose, as [described here](https://docs.alephium.org/frequently-asked-questions/#why-can-the-dapp-only-connect-to-one-of-my-addresses) -- that's probably something crucial that multi-chain dapp or multichain-wallet developers will need a heads up about!

alephium/caip10.md
bumblefudge: Actually, I realized that `/` won't work in a CAIP-2 reference, the only diacritics allowed are `-` and `_`.  Since `-1` makes sense as the wildcard, that just leaves `_` as the discriminant.

Wouldn't it make more sense, given all this, to use this as the syntax?

`alephium + ":" + group { -1 | 0 | 1 | 2 | 3} + "_" + network_id `

bumblefudge:
> 2. For address, we specify `alephium:${network_id}:${address}`. Because `address` already have the group information implicitly we don't need to specify it.

Actually, I would argue it's still worth encoding group in the CAIP-2 portion of a CAIP-10 because the group information requires quite a bit of sniffing, decoding and transformation to _extract_, whereas the counterparty handed a CAIP-10 may not have the compute to spare and would probably only do that much compute to _verify_ what they most often can just trust from the counterparty...

Lots of multichain stuff is kind of verbose or redundant in this way, to spare all parties in the system to keep all the tooling from every ecosystem on hand :D

alephium/caip19.md
h0ngcha0: Thanks! I added the special case for `universal` in this [commit](https://github.com/ChainAgnostic/namespaces/pull/117/commits/d128daedf783e31eef823ee20259c85fd77977f4)

alephium/caip2.md
h0ngcha0: Thanks! Added some explanation according to your suggestion [here](https://github.com/ChainAgnostic/namespaces/pull/117/commits/d128daedf783e31eef823ee20259c85fd77977f4), do you think it is more clear this way?

alephium/caip10.md
h0ngcha0: Thanks! I have updated the chain id format in this [commit](https://github.com/ChainAgnostic/namespaces/pull/117/commits/b16b4fe480accc5bde28f5134e70f9aabf0e9bb0)

h0ngcha0:
> > 2. For address, we specify `alephium:${network_id}:${address}`. Because `address` already have the group information implicitly we don't need to specify it.
> 
> Actually, I would argue it's still worth encoding group in the CAIP-2 portion of a CAIP-10 because the group information requires quite a bit of sniffing, decoding and transformation to _extract_, whereas the counterparty handed a CAIP-10 may not have the compute to spare and would probably only do that much compute to _verify_ what they most often can just trust from the counterparty...
> 
> Lots of multichain stuff is kind of verbose or redundant in this way, to spare all parties in the system to keep all the tooling from every ecosystem on hand :D

Thanks, I have made the update in this [commit](https://github.com/ChainAgnostic/namespaces/pull/117/commits/b16b4fe480accc5bde28f5134e70f9aabf0e9bb0)

bumblefudge (approved):
LGTM, waiting on 2nd editor approval or 2 independent approvals

h0ngcha0:
> LGTM, waiting on 2nd editor approval or 2 independent approvals

Thanks a lot @bumblefudge for your time and suggestions!

Hey @obstropolos , would really appreciate it if you can take a look at this PR when you have time, thanks very much :pray:

alephium/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Namespace for Alephium Blockchains
+
+This document defines the applicability of CAIP schemes to the
+blockchains of the Alephium ecosystem.
+
+Alephium uses the stateful UTXO model where assets are managed by
+UTXOs while contract states are managed using account-based
+model. Alephium is also a sharded blockchain, its state is divided
+into groups, with seperate chains responsible for processing
```
obstropolos, line 20:
```suggestion
into groups, with separate chains responsible for processing
```

alephium/caip2.md
```diff
@@ -0,0 +1,78 @@
...
+This is the implementation of CAIP-2 for Alephium network.
+
+## Rationale
+
+Alephium is a sharded blockchain that organizes addresses and states
+into distinct groups. The blockchain id includes both the network id
+and the group information.
+
+Each transaction has an origin and destination group. All transactions
+from by addressses in group `G1` to addresses in group `G2` will be
```
obstropolos, line 24:
```suggestion
from by addresses in group `G1` to addresses in group `G2` will be
```

obstropolos (changes requested):
Small errata - will approve once corrections are made. 

Thanks for putting this together!

h0ngcha0:
> Small errata - will approve once corrections are made.
> 
> Thanks for putting this together!

Both suggestions applied, thanks for your time!

obstropolos (approved):

PR #119, opened 2024-07-16
Proposed namespace: Wallet

### LINKED ISSUE #55: Special case: A distinct namespace for offline/wallet capabilities?
1.) DIDs and VCs are examples of very blockchain-like capabilities that a blockchain wallet might want to support, without requiring any communications with nodes.  Currently, CAIP-169 tries to update some work done by Mircea and Oliver at Veramo/Mesh to bring VC protocols into the toolkit of Wallet-Connect wallets.  These are namespace-independent, and entirely node-free.  So where do dapps put these methods when calling CAIP-25 on the wallet? A scope object called `wallet`, I'm thinking.  

2.) I'm presuming there are other offchain capabilities (PassKey comes to mind, or FIDO, or on-device biometrics) that it might behoove us to abstract and standardize some day!  These would also go in the `wallet` scope space, right?

2.) Note, in the evolution of the EVM RPC methods, `wallet_changeEthereumChain` and a few other `wallet_`-prefixed methods have been defined by EIPs.  These should probably live in `eip155`, confusingly enough, since `wallet::changeEthereumChain` doesn't make much sense for the non-EVM world!

Please chime in if you have objections or use-cases or to bikeshed the name `wallet` versus `offchain` versus `nodefree`

bumblefudge:
Rather than hardcode a special-case into CAIP-25, I would propose hard-coding an "abstraction" namespace for the special-case of subsetting other namespaces (for the purposes of, say, enabling an app<>wallet connection or app<>wallet RPC methods from a given L1 without needing to request and receive a network connection for that L1).

@Vandan @jiexi @gudahtt for discussion next week-- feel free to commit or PR against the branch if you have ideas, changes, examples for the "## Test Cases" section, and/or considerations about privacy, security, transport diversity, etc. Maybe this is a helpful starting point for other conversations around post-provider-injection transports?

Proposed as an alternative approach to [CAIPs#303](https://github.com/ChainAgnostic/CAIPs/pull/303)

also closes #55 if merged.


adonesky1 (approved):
Seems like the right idea!

wallet/caip2.md
```diff
@@ -0,0 +1,54 @@
...
+This combination (i.e. `wallet:eip155`) refers not to an `eip155` network but specifically to an application<>wallet connection defined according to the wallet and RPC conventions of the `eip155` namespace.
+
+## Rationale
+
+Implementers of [CAIP-25][] and other multichain application<>wallet connections have long struggled with how to map offchain or direct-to-wallet methods, notifications, and metadata to one another across namespaces, and how to safely and ergonomically segregate them from node-dependent/network-dependent functions.
+For this reason, an abstraction (in the form of a "meta-namespace" routing to the other namespaces) is here proposed.
+
+## Semantics
+
+`wallet:{namespace}` is functionally an alias for `{namespace}:wallet`, where `:wallet` refers to a special-case of a wallet connection, not a proper network (as required by the CAIP-2 profile for `eip155`).
```
jiexi, line 26: sorry for late review. My understanding of introducing this special `wallet` meta namespace was so that we could avoid the usage of `:wallet` entirely. This sentence seems to imply that `{namespace}:wallet` is also valid and interchangeable with `wallet:{namespace}`. Perhaps I am reading this wrong?

adonesky1: Yeah I agree this is a bit confusingly phrased and sortof implies (at least to my ear) that `{namespace}:wallet` is an established pattern that we're providing an alternative syntax for...

PR #126, opened 2024-08-19
Remove Redundant Chain ID Definition for Alephium

h0ngcha0:


bumblefudge (approved):
whoops embarassing that that slipped by me last time!

Uncontroversial and editorial, so skipping second review in this case.

PR #128, opened 2024-09-12
Fix markdown links

frangio:
These links are currently broken on GitHub and [on the website](https://namespaces.chainagnostic.org/cosmos/caip2).


bumblefudge:
Thanks, good catch!

PR #139, opened 2025-03-04
update syntax and example after 5792 syntax changes

bumblefudge:
Updated the 5792 `wallet_getCapabilities` equivalence spec language and examples following the introduction of the ["chainId 0" convention](https://github.com/eth-infinitism/EIPs/commit/5f3a1cdd91b88e9bb6d4b018dfa940bbe9f94961#diff-7a494500c37dc4b57613207d62874322198a4ced215f4a0457a0af908bb6364eR280) in Alex Forshtat's PR on 5792 [here](https://github.com/ethereum/EIPs/pull/8826)

### DIFF
--- a/eip155/caip25.md
+++ b/eip155/caip25.md
@@ -34,7 +34,10 @@ see the [eip155/caip211.md](./caip211.md) profile for further guidance on using 
 No namespace-wide or network-specific session properties have yet been proposed for standardization.
 When crafting such properties for contextual/in-network usage, it is recommended to align one's semantics and syntax (including case-sensitive style guides for property names!) with the [EIP-6963] wallet provider interface (which extends the [EIP-1193] interface) for common properties across architectures, making sure to avoid any collisions.
 
-Similarly, wherever possible, session properties should align closely with information passed as JSON objects by `wallet_` RPC methods, like capabilities or permissions. For example, the capability objects keyed to hexidecimal [EIP-155] `chainId`s defined as the expected return content of the `wallet_getCapabilities` method in [EIP-5792] should also be valid keyed the same way in `sessionProperties.capabilities`.
+Similarly, wherever possible, session properties should align closely with information passed as JSON objects by `wallet_` RPC methods, like capabilities or permissions. 
+For example, the capability objects specified in [EIP-5792] get passed via the `wallet_getCapabilities` RPC method inside of partition objects keyed to the hexadecimal chainId when chain-specific, and inside of an object keyed `0x00` (following the chainId 0 convention; see the [`eip155`/caip2 profile](caip2)) when universal to the `eip155` namespace.
+Should an application request and a wallet choose to pre-declare at time of connection, both parties can put such objects in the appropriately-keyed `scopedProperties` partition for chain-specific capabilities and in the `sessionProperties` (unpartitioned) for `eip155`-wide capabilities.
+See the example below, equivalent to the illustrative examples in [EIP-5792].
 
 ## Examples
 
@@ -52,7 +55,7 @@ Similarly, wherever possible, session properties should align closely with infor
         "methods": ["eth_sendTransaction", "eth_signTransaction", "eth_sign", "get_balance", "personal_sign"],
         "notifications": ["accountsChanged", "chainChanged"]
       },
-      "eip155:10": {
+      "eip155:8453": {
         "methods": ["get_balance"],
         "notifications": ["accountsChanged", "chainChanged"]
       },
@@ -65,16 +68,34 @@ Similarly, wherever possible, session properties should align closely with infor
       }
     },
     "optionalScopes":{
-      "eip155:42161": {
+      "eip155:84532": {
         "methods": ["eth_sendTransaction", "eth_signTransaction", "get_balance", "personal_sign"],
         "notifications": ["accountsChanged", "chainChanged"]
     },
     "sessionProperties": {
       "expiry": "2022-12-24T17:07:31+00:00",
       "caip154": {
-          "supported":"true"
+        "supported":"true"
+      },
+      "flow-control": {
+        "supported": "true"
       }
-    }         
+    },
+    "scopedProperties": {
+      "8453": {
+        "paymasterService": {
+          "supported": true
+        },
+        "sessionKeys": {
+          "supported": true
+        }
+      },
+      "84532": {
+        "auxiliaryFunds": {
+          "supported": true
+        }
+      }
+    }
   }
 }
 ```
@@ -94,15 +115,10 @@ Similarly, wherever possible, session properties should align closely with infor
         "notifications": ["accountsChanged", "chainChanged"],
         "accounts": ["eip155:1:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb", "eip155:137:0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb"]
       },
-      "eip155:10": {
-        "methods": ["get_balance"],
-        "notifications": ["accountsChanged", "chainChanged"],
-        "accounts:" []
-      },
-      "eip155:42161": {
+      "eip155:83532": {
         "methods": ["personal_sign"],
         "notifications": ["accountsChanged", "chainChanged"],
-        "accounts":["eip155:42161:0x0910e12C68d02B561a34569E1367c9AAb42bd810"]
+        "accounts":["eip155:83532:0x0910e12C68d02B561a34569E1367c9AAb42bd810"]
       },
       "wallet": {
         "methods": ["wallet_getPermissions", "wallet_switchEthereumChain", "wallet_getCapabilities"],
@@ -113,23 +129,20 @@ Similarly, wherever possible, session properties should align closely with infor
       }
     },      
     "sessionProperties": {
-      "capabilities": {
-          "0xab16a96d359ec26a11e2c2b3d8f8b8942d5bfcdb": {
-            "0x31": {
-              "paymasterService": {
-                 "supported": true
-               },
-               "sessionKeys": {
-                  "supported": true
-               }
-            },
-            "0x3432313631": {
-              "sessionKeys": {
-                "supported": false
-              }              
-            }
-          }
-       }   
+      "expiry": "2022-12-24T17:07:31+00:00",
+      "caip154": {
+        "supported":"true"
+      },
+      "flow-control": {
+        "supported": "true"
+      }
+    },
+    "scopedProperties": {
+      "84532": {
+        "auxiliaryFunds": {
+          "supported": false
+        }
+      }
     }
   }
 }


eip155/caip25.md
```diff
@@ -113,23 +129,20 @@ Similarly, wherever possible, session properties should align closely with infor
...
-            }
-          }
-       }   
+      "expiry": "2022-12-24T17:07:31+00:00",
+      "caip154": {
+        "supported":"true"
+      },
+      "flow-control": {
+        "supported": "true"
+      }
```
jakubuid, line 138: Would it be more beneficial to define the flow-control capabilities already in here instead of just sending a flag?

E.g.:
- flow-control in sessionProperties defined globally for all sessionScopes:
```
"sessionProperties": {
	"flow-control": {
		 "loose": ["halt", "continue"],
          	"strict": ["continue"]
	}
}
```

- flow-control in scopedProperties defined per chain:
```
"scopedProperties": {
	"84532"{
		"flow-control": {
		 	"loose": ["halt", "continue"],
          		"strict": ["continue"]
		}
	}
}
```

It would reduce round trips between the peers

Does it make sense to you all?

bumblefudge: Oh, I was just following exactly what's in the example in today's 5792; i thought the exact options for flow-control (and even dash versus camelcase) were still being debated OOB!

jakubuid: Agree on the example, looking good. This way peers can already exchange the flow-control capabilities aka we're reducing a need to call `wallet_getCapabilities`

bumblefudge: update [in this commit](https://github.com/ChainAgnostic/namespaces/pull/139/commits/37dc500ac61d2dd4d87fcc07f05b58d7a7cc471d) to use the selective example above

eip155/caip25.md
```diff
@@ -65,16 +68,36 @@ Similarly, wherever possible, session properties should align closely with infor
...
+        "supported":"true"
+      },
+      "flow-control": {
+        "loose": [], 
+        "strict": [], 
+        "exoticThirdThing": [] //caller is requesting a configuration-set which the wallet will drop as unrecognized      
+      }
+    },
+    "scopedProperties": {
+      "8453": {
```
llbartekll, line 87: why not caip2?

bumblefudge: whoops that was a mistake on my part, you're right that it should match the examples in the CAIP-25 spec and use full CAIP-2s

bumblefudge:
@llbartekll @jakubuid @pedrouid + anyone else following 5792 prototyping at ReOwn - approvey approvey before 5792 goes to last call so I can update the commit-specific link IN EIP-5792?

jakubuid:
> @llbartekll @jakubuid @pedrouid + anyone else following 5792 prototyping at ReOwn - approvey approvey before 5792 goes to last call so I can update the commit-specific link IN EIP-5792?

@bumblefudge before merging, according to the last 5792 updates (https://github.com/ethereum/EIPs/pull/9529/files#diff-7a494500c37dc4b57613207d62874322198a4ced215f4a0457a0af908bb6364eR353) can we add example of `"atomic": true` being sent in sessionProperties/scopedProperties?

bumblefudge:
OK i just added it in the commit above to match the open PR you linked to, although I'm still confused why the other capabilties are:
` { "supported": true }` 
and atomic is
` true `
! I opened a "change request" on that PR to be notified if it was just an oversight and the author updates the PR

llbartekll:
> > > @llbartekll @jakubuid @pedrouid + anyone else following 5792 prototyping at ReOwn - approvey approvey before 5792 goes to last call so I can update the commit-specific link IN EIP-5792?
> > 
> > 
> > @bumblefudge before merging, according to the last 5792 updates (https://github.com/ethereum/EIPs/pull/9529/files#diff-7a494500c37dc4b57613207d62874322198a4ced215f4a0457a0af908bb6364eR353) can we add example of `"atomic": true` being sent in sessionProperties/scopedProperties?
> 
> E.g. in scopedProperties:
> 
> ```
> caip2:{
>        "atomic":{
>            "supported":true
>         }
>    }
> ```
> 
> CC @llbartekll

@bumblefudge 
should this be allowed?:

```
caip2:{
    caip10A: {
       "atomic":{
           "supported":true
        }
     }
   caip10B: {
       "atomic":{
           "supported":false
        }
     }
   }
```

bumblefudge:
> should this be allowed?:

lol no that was a typo on my part, thank you for not making me feel more embarassed than I need to

jakubuid:
> > should this be allowed?:
> 
> lol no that was a typo on my part, thank you for not making me feel more embarassed than I need to

the point is that we need an adrress (caip10), because wallets can have many addresses on given chain and some of them could support atomic batches and the other not

bumblefudge:
> we need an address (caip10), because wallets can have many addresses on given chain and some of them could support atomic batches and the others not

Oh sorry, I missed that part.  EIP-5792 only allows this to be set per-chain, and leaves it up to the wallet to manage which addresses on that chain can do it, and whether or not "from" address in the batch is one it controls (and is an onchain-address). so maybe the solution here is to leave this complexity OUT of the examples in the CAIP-25 profile, but include a solution/workaround in the WC docs, e.g. multiple namespaces for the same chain where one returns true and one returns false for atomicity?

jakubuid:
> > we need an address (caip10), because wallets can have many addresses on given chain and some of them could support atomic batches and the others not
> 
> Oh sorry, I missed that part. EIP-5792 only allows this to be set per-chain, and leaves it up to the wallet to manage which addresses on that chain can do it, and whether or not "from" address in the batch is one it controls (and is an onchain-address). so maybe the solution here is to leave this complexity OUT of the examples in the CAIP-25 profile, but include a solution/workaround in the WC docs, e.g. multiple namespaces for the same chain (with different addresses) where one returns true and one returns false for atomicity?

The 5792 wallet_getCapabilities request allows a dapp to specify the address on which it needs the capabilities: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md#wallet_getcapabilities-example-parameters

In CAIP-25 handshake, the wallet can define 5792 capabilities in front for all of the addresses it supports, hence the thinking is that scopedProperties can be used for that as @llbartekll suggested:

```
caip2:{
    caip10A: {
       "atomic":{
           "supported":true
        }
     }
   caip10B: {
       "atomic":{
           "supported":false
        }
     }
   }
```
It should be convenient for wallets to do it this way

> multiple namespaces for the same chain (with different addresses) where one returns true and one returns false for atomicity?
I'm not sure if I follow how that structure would look like but I'm open for any suggestions

bumblefudge:
oh god you're totally right, i forgot that the REQUEST specifies a wallet, and optionally chains, and the response specifies chains (with the per-address implicit).  yeah modeling this inside CAIP-25 (where addresses are properties of namespaces) is... always going to be a little lossy/chaotic.  forgive me for complicating things but I think the EIP5792 exchange
```
req:
getCaps( address1, [chainX, chainY])
res:
chainX: {
   <caps>
}, chainY: {
   <caps>
}
```
as a CAIP-25 handshake would be more accurate to translate as:
```
req:
...
eip155:chainX: {
...
eip155:chainY: {
...
sessionProperties: {
   <caps>
}
res:
eip155:chainX: {
  addresses: [address1]
...
eip155:chainY: {
  addresses: [address1]
...
scopedProperties: {
   eip155:chainX: {
      <caps>
   }
}
```
(risky/potentially broken if multiple addresses are being requested) or
```
req:
...
eip155:chainX: {
...
eip155:chainY: {
...
sessionProperties: {
   <caps>
}
res:
eip155:chainX: {
  addresses: [address1]
...
eip155:chainY: {
  addresses: [address1]
...
scopedProperties: {
   "eip155:chainX": {
      "eip:155:chainX:address1": {
          <caps>
      }
   }
}
```
(more faithfully roundtrip-translateable, less risky of malforming, but extremely verbose).

although it's repetitive, I actually prefer the second of the two options I just gave, because the capabilities are being expressed PER ADDRESS inside of a namespace object (and repeated for every chain that address operates on, unless it's a 0x00/sessionProperties cap available on any eip155).  that is, at least, the functional complete-equivalent of the EIP5792 model.  i don't like the idea of just throwing CAIP-10 partitions into scopedProperties, because each entry in sessionProperties is specified to map to a scope (i.e. a "namespace" in Reown SDK terms) and some implementations (like I believe MM's) would just drop them if they were keyed to anything else. does that make sense?

llbartekll:
> oh god you're totally right, i forgot that the REQUEST specifies a wallet, and optionally chains, and the response specifies chains (with the per-address implicit). yeah modeling this inside CAIP-25 (where addresses are properties of namespaces) is... always going to be a little lossy/chaotic. forgive me for complicating things but I think the EIP5792 exchange
> 
> ```
> req:
> getCaps( address1, [chainX, chainY])
> res:
> chainX: {
>    <caps>
> }, chainY: {
>    <caps>
> }
> ```
> 
> as a CAIP-25 handshake would be more accurate to translate as:
> 
> ```
> req:
> ...
> "eip155:chainX": {
> ...
> "eip155:chainY": {
> ...
> "sessionProperties": {
>    <caps>
> }
> res:
> eip155:chainX: {
>   addresses: [address1]
> ...
> eip155:chainY: {
>   addresses: [address1]
> ...
> scopedProperties: {
>    eip155:chainX: {
>       <caps>
>    }
> }
> ```
> 
> (risky/potentially broken if multiple addresses are being requested) or
> 
> ```
> req:
> ...
> "eip155:chainX": {
> ...
> "eip155:chainY": {
> ...
> "sessionProperties": {
>    <caps>
> }
> res:
> eip155:chainX: {
>   addresses: [address1]
> ...
> eip155:chainY: {
>   addresses: [address1]
> ...
> scopedProperties: {
>    "eip155:chainX": {
>       "eip:155:chainX:address1": {
>           <caps>
>       }
>    }
> }
> ```
> 
> (more faithfully roundtrip-translateable, less risky of malforming, but extremely verbose).
> 
> although it's repetitive, I actually prefer the second of the two options I just gave, because the capabilities are being expressed PER ADDRESS inside of a namespace object (and repeated for every chain that address operates on, unless it's a 0x00/sessionProperties cap available on any eip155). that is, at least, the functional complete-equivalent of the EIP5792 model. i don't like the idea of just throwing CAIP-10 partitions into scopedProperties, because each entry in sessionProperties is specified to map to a scope (i.e. a "namespace" in Reown SDK terms) and some implementations (like I believe MM's) would just drop them if they were keyed to anything else. does that make sense?

maybe 2 options should be allowed?
option 1 if wallet supports given capabilities for all the addresses and option 2 for granular approach.

jakubuid:
@llbartekll @bumblefudge Should we update the example with the 2nd approach and merge this week?

llbartekll (approved):

llbartekll (commented):
lgtm

cryptotavares (commented):
Looks good to me, just wondering if you want to include the changes just added to [5792](https://github.com/ethereum/EIPs/pull/9601/files)

eip155/caip25.md
```diff
@@ -34,7 +34,11 @@ see the [eip155/caip211.md](./caip211.md) profile for further guidance on using
...
+Similarly, wherever possible, session properties should align closely with information passed as JSON objects by `wallet_` RPC methods, like capabilities or permissions. 
+For example, the capability objects specified in [EIP-5792] get requested per-address via the `wallet_getCapabilities` RPC method, and the returned objects are partitioned by the hexadecimal chainId when chain-specific, and inside of an object keyed `0x00` (following the chainId 0 convention; see the [`eip155`/caip2 profile](caip2)) when universal to the `eip155` namespace.
+Should an application request these capabilities and a wallet choose to pre-declare at time of connection, both parties can put such objects in the appropriately-keyed `scopedProperties` partition for chain-specific capabilities and in the `sessionProperties` (unpartitioned) for `eip155`-wide capabilities;
```
adonesky1, line 39: > and in the `sessionProperties` (unpartitioned) for `eip155`-wide capabilities;

Hmmm I'm realizing that my gut feeling is that `sessionProperties` should be reserved for properties that live fully outside of any given namespace and that `scopedProperties` should allow you to map things to a namespace **or** a full CAIP-2 chainId scope.

bumblefudge: yeah, i feel you, but full CAIP-2 scopes are out, remember? hehe

adonesky1: > full CAIP-2 scopes are out, remember
in what context? They're clearly still the keys in `scopedProperties`?

bumblefudge:
> Looks good to me, just wondering if you want to include the changes just added to [5792](https://github.com/ethereum/EIPs/pull/9601/files)

oh man it never ends! does [this](https://github.com/ChainAgnostic/namespaces/pull/139/commits/b43ffb58ff3b428d0f9215237369f0ce74b1f9e2) work? it's a little far-fetched as an example, but i wanted the reader to notice the heirarchy/override stuff.

jakubuid:
Hey folks, can we merge it shortly? What's missing?

eip155/caip25.md
bumblefudge: doh i meant full-namespace scopes

bumblefudge:
Approved out-of-band by @obstropolos

PR #140, opened 2025-04-23
Caip 350 namespaces for `solana` and `eip155`

0xteddybear:


eip155/caip350.md
```diff
@@ -0,0 +1,79 @@
...
+## Chain reference
+
+### Text representation
+```
+eip155:<number>
+```
+Where `<number>` is the decimal representation of the chain's `chainid`, without leading zeroes.
+
+##### Text representation -> CAIP-2 conversion
+In the case where the `chainid` is larger than what can be represented in 32 decimal characters, the leading 32 characters should be used
```
bumblefudge, line 24:
```suggestion

In the case where the `chainid` is larger than what can be represented in 32 hexidecimal characters, the leading 32 characters should be used
```

solana/caip350.md
```diff
@@ -0,0 +1,75 @@
...
+CAIP-2 namespace: `solana`
+
+## Chain reference
+Solana networks are distinguished by its genesis blockhash, which is normally represented as 44 base58btc characters, corresponding to 32 bytes. We chose to use the blockhash in full, as opposed to CAIP-2, for consistency with `eip155` and to avoid the complexity of slicing base58btc-encoded text [^1].
+
+[^1]: CAIP-2 limits chain references to 32 characters, so the `solana` namespace instructs to only use the first 32 characters of the base58btc-encoded genesis blockhash.
+Truncating the base58btc-encoded representation instead of the binary data, however, means the binary representation will differ significantly from what the Solana node might store in its own memory. e.g while truncating the text yields a string representing a 23-byte long result, it does not correspond to simply slicing the first 23 bytes from the genesis blockhash.
+
+### Text representation
+The full base58btc-encoded genesis blockhash used used. This is larger than the CAIP-2 representation.
```
bumblefudge, line 22:
```suggestion
The full base58btc-encoded genesis blockhash is used. This is larger than the CAIP-2 representation.
```

bumblefudge:
This makes sense for the most part, i have only two minor comments:

1. blank lines after headers! (i can always lint/style-guide-conform this PR at the last minute before merge if you don't have time)
2. doesn't Coinbase already use ["hexadecimal chainIds"](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5792.md?plain=1#L100) in their RPCs, such as in EIP-5792? you define it from scratch here, and I would check if it isn't already defined somewhere just in case there's some corner-case where this logic and whatever coinbase does would create [slightly] different bytestrings?

eip155/caip350.md
0xteddybear: :eyes: I believe decimal is the correct word, since eip155's CAIP-2 profile defines an example using auxilium network mainnet, with chainid 28945486 (decimal, I checked :grin: )

0xteddybear:
> doesn't Coinbase already use "hexadecimal chainIds" in their RPCs, such as in EIP-5792? you define it from scratch here, and I would check if it isn't already defined somewhere just in case there's some corner-case where this logic and whatever coinbase does would create [slightly] different bytestrings?

EIP-5792 requires the 0x-prefixed 'hexadecimal' (no specification on capitalization) string to be used, which AFAICT does not conflict with the formats defined in this PR, since:
- the Interoperable Name format is base10, similar to CAIP-2
- the Interoperable Address' text representation is not a base16 string, but the Interoperable Name, so there will be no need to display the former as a 'bytestring'

solana/caip350.md
```diff
@@ -0,0 +1,93 @@
...
+
+The full base58btc-encoded genesis blockhash is used. This is larger than the CAIP-2 representation.
+
+##### Text representation -> CAIP-2 conversion
+
+The leading 32 characters are used, and the rest discarded.
+
+##### CAIP-2 - text representation conversion
+
+This transformation is not fully deterministic. It is assumed wallets and other software will be able to differentiate between chains with just the leading 32 base58btc-encoded characters, and use a lookup table of popular chains to complete the missing information to convert CAIP-2 identifiers to this standard.
```
bumblefudge, line 33:
```suggestion
This transformation is not fully deterministic. It is assumed wallets and other software will be able to differentiate between chains with just the leading 32 base58btc-encoded characters, and use a lookup table of chains to complete the missing information to convert CAIP-2 identifiers to this standard.
```
they don't need to be popular or even public, just expected/known.  private or ephemeral chainIds would also need to be in that lookup table/mapping, and even calling it a table makes it sound more static than it would have to be in some cases... but i think it's fine to say "if you use CAIP-2 you already have some kind of local lookup mechanism, which is still required here".  but what i think should be made clearer is whether CAIP-2 support is optional or required, on the parsing/receiving side in particular (how does Postel's Law  apply? does a receiving implementation need to have such a lookup? or just be able to recognize/validate it as a CAIP-2 and throw a specific error that says "sorry i don't do CAIP-2" distinct from an "unknown chain" error?)

bumblefudge:
Skipping second approval to unblock EIP merging downstream.

PR #150, opened 2025-08-18
Add iota namespace

UMR1352:
Implementation of CAIP-2 for the IOTA's namespace.
PRs for CAIP-10 and CAIP-19 are on their way too.


bumblefudge (approved):
looks good to me!

obstropolos (approved):

PR #151, opened 2025-08-28
IOTA CAIP-10

UMR1352:
Adds an implementation of CAIP-10 for the iota namespace

### DIFF
--- a/iota/caip10.md
+++ b/iota/caip10.md
@@ -1,0 +1,210 @@
+---
+namespace-identifier: iota-caip10
+title: IOTA - Account ID Specification
+author: Enrico Marconi (@UMR1352)
+status: Draft
+type: Informational
+created: 2025-07-28
+requires: CAIP-2, CAIP-10
+---
+
+<!--You can leave these HTML comments in your merged CAIP and delete the 
+ visible duplicate text guides, they will not appear and may be helpful to 
+ refer to if you edit it again. This is the suggested template for new CAIPs.
+ Note that an CAIP number will be assigned by an editor. When opening a pull
+ request to submit your EIP, please use an abbreviated title in the 
+ filename, `caipX.md`, all lowercase, no `-` between the CAIP and its 
+ number.-->
+
+# CAIP-10
+
+*For context, see the [CAIP-10] specification.*
+
+## Introduction 
+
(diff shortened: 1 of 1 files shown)


iota/caip10.md
```diff
@@ -0,0 +1,210 @@
...
+validated against the chain they are targeting or depending to to avoid replay
+attacks or other unintended outcomes. This is often done by an API or RPC call
+to a node to validate the targetted chain or network. Include a sample
+request/response and add the relevant documentation to the `## References`
+section below if possible, as well as an explanation of any steps needed to
+validate the results, calculate checksums, persist session metadata or nonces,
+ etc. -->
+
+Any Account ID, conforming to the aforementioned syntax is a valid IOTA Account ID,
+indipendently of whether it is already in use by a wallet or on-chain object.
```
bumblefudge, line 86:
```suggestion
independently of whether it is already in use by a wallet or on-chain object.
```

iota/caip10.md
```diff
@@ -0,0 +1,210 @@
...
+an object, but a key-derived address.
+
+## Rationale
+
+<!-- Explain here how the mapping or translation between native identifiers and
+CAIP-10 identifiers was arrived at, history and pre-history, etc.-->
+
+IOTA addresses are lowercase-hex encoded 32 bytes string, which in the case of key-pair derived addresses - i.e.
+traditional accounts - are computed by hashing with the `blake2b` algorithm a public key's bytes prepended by a
+single tag byte that identifies the type of key used - e.g. `ed25519`, `secp256k1`, etc.
```
bumblefudge, line 163:
```suggestion
IOTA address currently supports pure Ed25519, Secp256k1, Secp256r1, and MultiSig with corresponding flag bytes of 0x00 (not prepended), 0x01, 0x02, and 0x03, respectively.
```
This is just my best guess, it's actually unclear from [the docs](https://docs.iota.org/developer/cryptography/transaction-auth/keys-addresses#address-format), whether Ed25519 have a 0x00 byte prepended or no flag-byte prepended... maybe double-check before merging or rewording/refactoring my suggestion! but either way, you get the gist-- however it works, this is the level of detail that's most useful to include in the CASA profile.

Oh, and these docs should definitely be linked in the ## References section!

PR #178, opened 2026-03-18
Add `atto` namespace and CAIP-2 specification

rotilho:
This PR proposes a new `atto` namespace for the Atto ecosystem and adds an initial CAIP-2 profile for Atto chain identifiers.

Why this should be its own namespace:
- Atto is a native layer-1 payment network, not an environment of an existing namespace.
- It has a distinct account-chain model, native address format, native transaction format, and native proof-of-work validation rules.
- Tools interacting with Atto need Atto-specific validation logic.

### DIFF
--- a/atto/README.md
+++ b/atto/README.md
@@ -26,21 +26,27 @@ A dedicated namespace allows tools to refer to Atto deployment environments cons
 
 ## Governance
 
-This namespace tracks the Atto protocol and its network identifiers as defined by the Atto ecosystem maintainers.
-Future CAIP profiles for Atto should follow this namespace to document account IDs, asset IDs, and other ecosystem-specific conventions.
-Public network naming should remain stable over time to ensure long-term interoperability.
+This namespace tracks the public Atto protocol as documented by the ecosystem's maintainers.
+Protocol-level changes and governance-relevant decisions should be validated against public Atto materials, especially the whitepaper and official documentation, and against the implementation history in the public GitHub repositories.
+Future updates to this namespace should cite those sources when describing changes to network behavior, distribution, staking, or other community-facing protocol decisions.
 
 ## References
 
 - [Documentation][] - Official documentation for the Atto ecosystem.
+- [Whitepaper][] - Public protocol and governance context, including decentralization and representative voting.
 - [Node API Reference][] - Reference for the Atto node RPC and REST interfaces.
 - [Integration Guide][] - Overview of Atto components, integration model, and operational guidance.
-- [GitHub][] - Source code and development resources.
+- [Blog][] - Public announcements and writeups for major ecosystem and protocol changes.
+- [GitHub][] - Source code, issues, and pull requests for implementation history and concrete protocol updates.
+- [Node Repository][] - Reference implementation of the Atto node and consensus behavior.
 
 [Documentation]: https://atto.cash/docs
+[Whitepaper]: https://atto.cash/docs/whitepaper
 [Node API Reference]: https://atto.cash/api/node
 [Integration Guide]: https://atto.cash/docs/integration
+[Blog]: https://atto.cash/blog
 [GitHub]: https://github.com/attocash
+[Node Repository]: https://github.com/attocash/node
 
 ## Copyright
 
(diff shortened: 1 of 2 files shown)


atto/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Rationale
+
+Atto needs its own namespace because it introduces a distinct execution and validation model.
+Native account-chain semantics, address encoding, and transaction serialization do not fit under existing namespaces such as EVM, Solana, or Stellar.
+A dedicated namespace allows tools to refer to Atto deployment environments consistently using human-readable CAIP-2 identifiers.
+
+## Governance
+
+This namespace tracks the Atto protocol and its network identifiers as defined by the Atto ecosystem maintainers.
+Future CAIP profiles for Atto should follow this namespace to document account IDs, asset IDs, and other ecosystem-specific conventions.
```
bumblefudge: Actually, this^ part is implicit-- if your namespace PR gets accepted, any future PRs extending or updating it are assumed to recognize your "first write wins" for conflict resolution 😅 .  Usually, this section is for adding links to help the editors confirm if the ecosystem itself passed an improvement proposal, hard fork, etc when a later contributor wants to update it!  

For instance, I saw [this section](https://github.com/waldolemmer/fork-docs-attocash/blob/main/whitepaper.md#governance-and-decentralization), which is probably worth linking to from this section, but maybe also link to whatever github or other public forum concrete protocol updates are discussed in?  That way, if a future contributor wants to update your PR, we can check to validate any claims about community governance decisions that happened in the interim.

atto/caip2.md
```diff
@@ -0,0 +1,150 @@
...
+The `atto` namespace is intended for chains that use the native Atto protocol and validation model.
+A valid CAIP-2 identifier in this namespace refers to an Atto network that shares Atto's block model, address model,
+signing rules, and proof-of-work rules.
+
+The following references are defined by this draft:
+
+- `live` — Atto mainnet
+- `beta` — Atto public beta / test network
+- `dev` — Atto development network
+- `local` — local or private development network
```
bumblefudge: Does this work like `localhost`?  I.e., dev is an actual network that I can write to and someone else will be able to reference, but `local` is assumed to be ephemeral and not published anywhere? it might be worth making this a little more explicit, as very few CAIP-2 profiles include this kind of "reserve word" with distinct semantics!

atto/caip2.md
```diff
@@ -0,0 +1,134 @@
...
+
+Suggested validation regex:
+
+```text
+^atto:(live|beta|dev|local)$
+```
+
+Implementations MUST treat the namespace component as exactly `atto`.
+Implementations SHOULD treat the reference component as lowercase.
+Unknown future references within the `atto` namespace SHOULD be rejected unless explicitly supported by the implementation.
```
bumblefudge, line 65: Does this mean the same private key will be able to sign for the same address on different chainIds, i.e., that it has an EVM-style replay vulnerability if a transaction is valid on multiple chains? 
If so, you might want a MUST not a SHOULD here (and a warning about replays and using a given key on multiple networks!) 😅

bumblefudge (approved):
Left two comments where optional clarifications might help, but otherwise good to go!

atto/caip2.md
rotilho: Yes, the same private key can be used to control the same address across different chain. However, transactions include the network ID in the signed payload so a transaction signed for one chain is not valid on another and cannot be replayed across chains.

Source: https://atto.cash/docs/integration/advanced/protocol-offline-signing-reference#overview-of-atto-blocks

atto/caip2.md
rotilho: Yep, exactly. `local` is meant to be ephemeral and local-only. We use it for local development, integration tests, and benchmarks; a fresh genesis is created on startup. `dev`, on the other hand, refers to an actual shared network that others can use.

Let me reword it to make that more explicit.

atto/README.md
rotilho: Right! :D

Rewrote this section and linked more resources in the References section

PR #182, opened 2026-04-06
Add acknacki namespace for Acki Nacki blockchain

Futurizt:
## Summary

Adds the `acknacki` namespace for the Acki Nacki blockchain — a probabilistic proof-of-stake consensus protocol optimized for fast finality with execution parallelization, running AVM (Advanced Virtual Machine) supporting TVM and WASM.

**Peer-reviewed publication:** [ACNS 2024 (Springer)](https://doi.org/10.1007/978-3-031-61486-6_4)

### Networks

| CAIP-2 | Network |
|--------|---------|
| `acknacki:0` | Mainnet |
| `acknacki:1` | Shellnet (testnet) |

### Files

- `acknacki/README.md` — Namespace overview, rationale, governance, Springer reference
- `acknacki/caip2.md` — Chain ID specification (`global_id` from block header)
- `acknacki/caip10.md` — Account ID specification (`DappID.AccountID`, 512-bit composite address)

### Key differentiators from existing namespaces

- **AVM**: supports both TVM and WASM execution (not TVM-only)
- **DApp ID**: unique addressing model — accounts identified by `DappID.AccountID` (not workchain:address)
- **Gasless internal transactions**: contracts within same DApp ID share gas pool via DappConfig
- **Probabilistic PoS consensus**: 2-step finality, subquadratic message complexity

### References

- Documentation: https://docs.ackinacki.com
- Developer portal: https://dev.ackinacki.com
- GitHub: https://github.com/ackinacki/ackinacki
- Explorer: https://mainnet.ackinacki.org

### DIFF
--- a/acknacki/README.md
+++ b/acknacki/README.md
@@ -14,11 +14,11 @@ consensus protocol optimized for fast finality, while allowing for high
 throughputs via execution parallelization. It achieves higher Byzantine
 fault tolerance than Nakamoto, BFT (including Hotstuff and AptosBFT),
 Solana, and other modern consensus protocols. The protocol reaches
-consensus in two communication steps and has a total number of messages
-that are subquadratic to the number of nodes.
+consensus in two communication steps and exchanges a total number of
+messages per consensus round that is subquadratic in the number of nodes.
 
-Acki Nacki runs AVM (Advanced Virtual Machine) supporting both TVM and
-WASM execution environments. It features a DApp ID system for gasless
+Acki Nacki runs AVM (Advanced Virtual Machine) supporting both Ton Virtual
+Machine (TVM) and WASM execution environments. It features a DApp ID system for gasless
 internal transactions, ECC (Extra Currency Collection) token transfers,
 and native support for AI agent operations.
 
@@ -37,9 +37,11 @@ A separate namespace from `tvm` is necessary because:
 
 ## Governance
 
-Acki Nacki protocol development is managed by GOSH Technology Ltd.
-Protocol specifications and node software are open source at
-[github.com/ackinacki/ackinacki][github].
+Acki Nacki core protocol developer is [GOSH](https://gosh.sh). Protocol
+specifications and node software are open source at
+[github.com/ackinacki/ackinacki][github]. Network updates are performed
+bitcoin-style — node owners update their nodes to a newest version at
+will. There is no DAO, foundation or governance body.
 
 The consensus protocol is formally described in a peer-reviewed
 publication at [ACNS 2024 (Springer)][springer].
--- a/acknacki/README.md
+++ b/acknacki/README.md
@@ -1,0 +1,65 @@
+---
+namespace-identifier: acknacki
+title: Acki Nacki Blockchain
+author: Mitja Goroshevsky (@Futurizt)
+status: Draft
+type: Informational
+created: 2026-04-07
+---
+
+# Namespace for Acki Nacki Blockchain
+
+Acki Nacki is a Rust implementation of a probabilistic proof-of-stake
+consensus protocol optimized for fast finality, while allowing for high
+throughputs via execution parallelization. It achieves higher Byzantine
+fault tolerance than Nakamoto, BFT (including Hotstuff and AptosBFT),
+Solana, and other modern consensus protocols. The protocol reaches
+consensus in two communication steps and has a total number of messages
+that are subquadratic to the number of nodes.
+
+Acki Nacki runs AVM (Advanced Virtual Machine) supporting both TVM and
+WASM execution environments. It features a DApp ID system for gasless
+internal transactions, ECC (Extra Currency Collection) token transfers,
+and native support for AI agent operations.
+
+## Rationale
+
+The `acknacki` namespace (8 characters, the CAIP-2 maximum) identifies
+the Acki Nacki blockchain family. The reference field uses the
+`global_id` integer from the block header, which uniquely identifies
+each network (mainnet, shellnet/testnet).
+
+A separate namespace from `tvm` is necessary because:
+- Acki Nacki runs AVM supporting TVM and WASM, evolving beyond TVM
+- The DApp ID threading model is unique to Acki Nacki
+- The consensus mechanism is specific to Acki Nacki
+- Address format uses DApp ID routing, not workchain IDs
+
+## Governance
+
+Acki Nacki protocol development is managed by GOSH Technology Ltd.
+Protocol specifications and node software are open source at
+[github.com/ackinacki/ackinacki][github].
+
+The consensus protocol is formally described in a peer-reviewed
+publication at [ACNS 2024 (Springer)][springer].
+
+## References
+
+- [Acki Nacki Documentation][docs] - Official documentation
+- [Acki Nacki Developer Portal][dev] - Developer guides and API reference
+- [Acki Nacki GitHub][github] - Node software and smart contracts
+- [Consensus Paper (Springer)][springer] - "Acki Nacki: A Probabilistic
+  Proof-of-Stake Consensus Protocol with Fast Finality and
+  Parallelisation" (ACNS 2024)
+- [Block Explorer][explorer] - Mainnet block explorer
+
+[docs]: https://docs.ackinacki.com
+[dev]: https://dev.ackinacki.com
+[github]: https://github.com/ackinacki/ackinacki
+[springer]: https://doi.org/10.1007/978-3-031-61486-6_4
+[explorer]: https://mainnet.ackinacki.org
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/acknacki/caip10.md
+++ b/acknacki/caip10.md
@@ -1,0 +1,148 @@
+---
+namespace-identifier: acknacki-caip10
+title: Acki Nacki Blockchain - Account ID Specification
+author: Mitja Goroshevsky (@Futurizt)
+discussions-to: https://github.com/ChainAgnostic/namespaces/pull/XXX
+status: Draft
+type: Standard
+created: 2026-04-07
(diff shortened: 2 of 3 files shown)


acknacki/README.md
```diff
@@ -0,0 +1,65 @@
...
+---
+
+# Namespace for Acki Nacki Blockchain
+
+Acki Nacki is a Rust implementation of a probabilistic proof-of-stake
+consensus protocol optimized for fast finality, while allowing for high
+throughputs via execution parallelization. It achieves higher Byzantine
+fault tolerance than Nakamoto, BFT (including Hotstuff and AptosBFT),
+Solana, and other modern consensus protocols. The protocol reaches
+consensus in two communication steps and has a total number of messages
```
bumblefudge: "total number of messages" is a little vague, maybe clarify? total per what?

acknacki/README.md
```diff
@@ -0,0 +1,65 @@
...
+that are subquadratic to the number of nodes.
+
+Acki Nacki runs AVM (Advanced Virtual Machine) supporting both TVM and
```
bumblefudge:
```suggestion
Acki Nacki runs AVM (Advanced Virtual Machine) supporting both Ton Virtual Machine (TVM) and
```
To contradistinguish from **Tron** VM, which uses the same acronym

acknacki/caip2.md
```diff
@@ -0,0 +1,110 @@
...
+
+# CAIP-2
+
+*For context, see the [CAIP-2][] specification.*
+
+## Introduction
+
+Acki Nacki is a layer-1 blockchain identified by a `global_id` integer stored
+in every block header. Each network (mainnet, shellnet) has a unique
+`global_id` assigned at genesis.
```
bumblefudge, line 20: Assigned by whom?

acknacki/caip2.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+The returned `global_id` value is the CAIP-2 reference for that network.
+
+## Rationale
+
+Using `global_id` from the block header follows the same pattern as the `tvm`
+namespace (which uses the identical field for Everscale and TON). This provides
+a simple, deterministic, on-chain verifiable identifier that requires no
+external registry.
```
bumblefudge, line 83: What happens if a new chain is started that "reuses" a chainId already in use? Or if a malicious network starts returning a different chainId over time when queried by the above method?  Good to explain this stuff (and link to further reading if "it's complicated") for first-timers, who are the target audience of this document.

acknacki/caip10.md
```diff
@@ -0,0 +1,148 @@
...
+*For context, see the [CAIP-10][] specification.*
+
+## Introduction
+
+Acki Nacki uses a two-part account addressing scheme. Each account is
+identified by a **DApp ID** (256 bits) and an **Account ID** (256 bits),
+forming a 512-bit composite address called `AccountRouting`.
+
+The DApp ID groups contracts into a logical application. All contracts
+within the same DApp ID share a gas pool (DappConfig) and can transact
```
bumblefudge:
```suggestion
within the same DApp ID share ONE (1) gas pool (defined in DappConfig) and can transact
```
(I'm just guessing here-- if I'm wrong, rephrase to be equally explicit but more accurate!)

acknacki/caip10.md
```diff
@@ -0,0 +1,148 @@
...
+- **DApp ID**: 256-bit identifier (32 bytes, 64 hex chars)
+- **Account ID**: 256-bit identifier (32 bytes, 64 hex chars)
+
+Together they form an `AccountRouting` — the full address used for
+message routing and thread assignment in the Acki Nacki network.
+
+### Syntax
+
+The CAIP-10 account address is the concatenation of the lowercase
+hex-encoded DApp ID and Account ID, separated by a period:
```
bumblefudge, line 69: I can tell from the examples below that hex-encoded DOESN'T use the `0x` prefix common in other blockchains, but you might want to explicitly call that out here for the lazy EVM-familiar skimmer who doesn't check the examples confronted with this ambiguity!

acknacki/caip10.md
```diff
@@ -0,0 +1,148 @@
...
+requires: CAIP-2, CAIP-10
+---
+
+# CAIP-10
+
+*For context, see the [CAIP-10][] specification.*
+
+## Introduction
+
+Acki Nacki uses a two-part account addressing scheme. Each account is
```
bumblefudge, line 18: I'm a little unclear on what an AccountID means by itself.  An "Externally owned account", for example, like a metamask on EVM: does that work by itself? Can a P2P transfer with no smart contracts involved work from a bare accountID to a bare accountID, and if so, who pays the gas?  

Regardless of whether all this is inherited from "vanilla" TVM, I would spell it out, because as-is it sounds like the only actors on Acki Nacki are what EVM calls "onchain accounts" (controlled/mediated via a deployed contract).

bumblefudge (changes requested):
Everything other than [this comment](https://github.com/ChainAgnostic/namespaces/pull/182/changes#r3160423365) is a minor stylistic suggestion, but I'd say it's ready to merge if this one blocking comment is addressed.

acknacki/README.md
Futurizt: Right, ambiguous. Rephrased to *"the total number of messages exchanged per consensus round is subquadratic in the number of nodes"* — that's what the Springer paper analyzes (per-round communication complexity). Done in 40434a1.

acknacki/README.md
Futurizt: Accepted, applied verbatim in 40434a1. Good distinction from Tron VM.

acknacki/caip2.md
Futurizt: There is no assigning authority — Acki Nacki is permissionless and decentralized. `global_id` is fixed at genesis and remains constant for the lifetime of the network. The public mainnet (`0`) and shellnet (`1`) values are conventionally reserved; anyone bootstrapping a private network selects their own value at genesis. Clarified in the Semantics section in 40434a1.

acknacki/caip2.md
Futurizt: Added a "Verification and Trust Model" subsection in 40434a1 covering both:

(a) protocol-level verification — `global_id` is embedded in every block header, so a client can cross-check the value against multiple Block Keepers rather than trusting a single GraphQL endpoint. Acki Nacki is permissionless, so no central registry can prevent two unrelated private networks from picking the same `global_id`; applications needing stronger guarantees should additionally fingerprint the chain (genesis block hash, well-known system contract addresses).

(b) for outside observers requiring cryptographic assurance of the chain history itself, Acki Nacki includes a network-history proof system based on zero-knowledge proofs that lets an external verifier validate the chain witness without trusting any single node. The proof system is currently under test and is planned for activation on mainnet to harden the witness for light clients and cross-chain verifiers.

acknacki/caip10.md
Futurizt: Correct — applied verbatim in 40434a1. Each DApp ID has exactly one `DappConfig` contract holding the shared SHELL pool that internal calls draw from via `gosh.mintshell()`.

acknacki/caip10.md
Futurizt: Added an explicit line in the Syntax section in 40434a1: *"Hex encoding is lowercase and does **not** include the `0x` prefix used in EVM-style addresses; both DApp ID and Account ID are exactly 64 hex characters with no separator."*

acknacki/caip10.md
Futurizt: Good catch — this is a genuine source of confusion for EVM-familiar readers. Added an "Account Model" subsection to caip10.md in 40434a1 spelling out:

- Acki Nacki uses an account-based model where **every account is itself a smart contract**; there are no EVM-style EOAs.
- A private key is not an actor on its own — it can only sign external messages addressed to an already-deployed contract whose code chooses whether to accept the signature and execute. A "user wallet" is therefore a deployed contract (typically a multisig).
- P2P value transfers always flow contract-to-contract:
  - **Inside a DApp ID** — gas is drawn from the shared `DappConfig` pool via `gosh.mintshell()` (gasless internal txs).
  - **From outside (external message)** — the receiving contract pays gas from its own balance. A bare AccountID with no deployed contract is uninitialized; incoming messages either bounce or, when accompanied by a `stateInit`, deploy the contract at that address.
  - **Cross-DApp** — native VMSHELL is zeroed at the protocol level; ECC tokens (e.g. SHELL) can cross.
- A bare AccountID is meaningful only as a destination address — it cannot originate a transaction until a contract has been deployed at it.

Futurizt:
Thanks for the thorough review! All seven inline comments addressed in 40434a1:

- **Blocking** (`caip10.md:18`): added an "Account Model" subsection making explicit that every account is a contract (no EOAs) and spelling out gas-payment rules per scope.
- Both suggestion blocks applied verbatim (TVM disambiguation, "ONE (1) gas pool" wording).
- "subquadratic messages" clarified to "per consensus round".
- `caip2.md` Semantics: dropped the "assigned by" framing — Acki Nacki is permissionless, no assigning authority; values are picked at genesis.
- New `caip2.md` "Verification and Trust Model" subsection covers BK cross-check, fingerprinting recommendation, and the in-test ZKP network-history proof system planned for mainnet activation.
- Hex-encoding convention spelled out (lowercase, no `0x` prefix).
- Also revised the README Governance section to accurately reflect the decentralized model (bitcoin-style updates by node owners; no DAO/foundation/governance body; GOSH as core protocol developer).

Ready for another pass.

Futurizt (commented):
Changes made as per resuest

bumblefudge (approved):
GTG, sorry for the delay between rounds, we check in monthly on these and often miss notifications in between!

Futurizt:
@obstropolos when you have a moment — bumblefudge approved 40434a1, would appreciate a second look, whenever you have a spare moment, thank you

obstropolos (approved):

PR #183, opened 2026-04-29
feat: add Keeta namespace

sc4l3r:
This PR adds a new `keeta` namespace for [Keeta](https://keeta.com), a high-performance Layer 1 blockchain built to meet the demands of the global financial system.

It includes specifications for CAIP2, CAIP10, and CAIP19.

## References

- Overview by Google Cloud: https://cloud.google.com/blog/topics/financial-services/how-blockchain-network-keeta-processes-11-million-transactions-per-second-with-spanner
- Product description: https://keeta.com/product-manual.pdf
- TypeScript SDK Docs: https://static.network.keeta.com/docs/index.html
- Explorer: https://explorer.keeta.com/ (official) or https://kee.tools/ (community)
- Wallet: https://wallet.keeta.com/ (mainnet) or https://wallet.test.keeta.com/ (testnet)

### DIFF
--- a/keeta/README.md
+++ b/keeta/README.md
@@ -1,0 +1,52 @@
+---
+namespace-identifier: keeta
+title: Keeta
+author: ["@sc4l3r"]
+discussions-to: https://github.com/ChainAgnostic/namespaces/pull/XXXX
+status: Draft
+type: Informational
+created: 2026-04-27
+requires: ["CAIP-2"]
+---
+
+# Namespace for Keeta
+
+[Keeta][keeta-home] is a Delegated Proof of Stake (DPoS) Layer-1 designed for high-throughput asset transfers, real-world asset tokenization, and built-in compliance primitives (KYC, sanctions, ACLs).
+The protocol uses a [Directed Acyclic Graph (DAG)][whitepaper] of per-account block chains rather than a single linearly-ordered ledger, which lets independent accounts publish concurrently and lets the network achieve sub-second finality.
+
+The `keeta` namespace describes both the public Keeta main network and any of its sibling networks -- the test, staging, and dev networks, and any privately-launched **sub-network** (a private instance that runs the same protocol but with isolated transaction visibility and its own validator set).
+All networks share the same key-pair format, address encoding, and operation set, so a single Keeta key pair can be used across networks and the only thing that distinguishes one from another in CAIP terms is its `networkId`.
+
+## Rationale
+
+Registering the `keeta` namespace enables standard CAIP-2 chain identifiers for every Keeta network distinguished by their numeric `networkId`.
+Keeta does not run an EVM or similar and has no smart-contract addresses.
+Therefore, CAIP profiles in this namespace focus on identifying networks, accounts (keyed and identifier-style), and tokens, not contracts.
+
+## Governance
+
+Keeta uses Delegated Proof of Stake.
+Token holders elect representatives, which vote on the validity of vote staples (atomic groups of blocks).
+Network-wide policy -- for example, the right to create new tokens or storage accounts -- is expressed as permissions on the [Network Account][accounts], a deterministically-generated identifier account that exists once per network.
+
+Protocol changes are coordinated by Keeta Token Genesis LLC and the validator set.
+The reference implementation is published as the [`@keetanetwork/keetanet-client`][sdk] TypeScript SDK; the protocol specification is described in the [Keeta whitepaper][whitepaper].
+
+## References
+
+- [Keeta home][keeta-home] - public site and ecosystem overview.
+- [Keeta whitepaper][whitepaper] - protocol specification, DAG model, consensus, and account model.
+- [Keeta documentation][keeta-docs] - developer guide, SDK reference, and anchor framework.
+- [Keeta SDK source][sdk] - the canonical reference implementation that defines `NetworkIDs`, address encoding, and key algorithms cited by the CAIP profiles in this namespace.
+- [Public network resources][official-links] - wallet, block explorer, and faucet endpoints for the public main and test networks.
+
+[keeta-home]: https://keeta.com/
+[whitepaper]: https://keeta.com/whitepaper.pdf
+[keeta-docs]: https://docs.keeta.com/
+[sdk]: https://github.com/KeetaNetwork/keetanet-client
+[official-links]: https://docs.keeta.com/other-documentation/official-links
+[accounts]: https://docs.keeta.com/components/accounts
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/keeta/caip10.md
+++ b/keeta/caip10.md
@@ -1,0 +1,190 @@
+---
+namespace-identifier: keeta-caip10
+title: Keeta - Account ID Specification
+author: ["@sc4l3r"]
+discussions-to: https://github.com/ChainAgnostic/namespaces/pull/XXXX
+status: Draft
+type: Standard
+created: 2026-04-27
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+_For context, see the [CAIP-10][] specification._
+
+## Introduction
+
+Every [account][accounts] on a Keeta network -- whether it is a key-pair-backed account (ECDSA secp256k1, ECDSA secp256r1, or Ed25519), a token, a storage account, a network account, or a multisig -- is addressed by a single base32-encoded **public key string** with the prefix `keeta_`.
+
+The body of the address self-identifies its account type via its first base32 characters (e.g. `aa` is a secp256k1 keyed account, `am` is a token, `aq` is a storage account, `ai` is a Network Account), and ends in a checksum.
+
+This profile uses the full canonical `keeta_...` form as the [CAIP-10][] `account_address`.
+
+## Specification
+
+### Semantics
+
+The inputs are:
+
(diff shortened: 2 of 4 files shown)


keeta/caip10.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+keeta:<networkId>:<address>
+```
+
+Where:
+
+- `keeta` is the namespace.
+- `<networkId>` is a CAIP-2 reference for a Keeta network as defined by the [Keeta CAIP-2 Profile][CAIP-2 Profile].
+- `<address>` is the canonical Keeta public key string, **including** its `keeta_` prefix.
```
sc4l3r, line 72: The `keeta_` prefix is up for discussion. Keeta addresses always include the prefix but according to [caip-10](https://standards.chainagnostic.org/CAIPs/caip-10) the `account_address` must follow `[-.%a-zA-Z0-9]{1,128}` which doesn't allow for `_`.

Would it be possible to add `_` to the allowed character set or should we strip the prefix?

bumblefudge: the former is essentially impossible, just do the latter!

keeta/README.md
```diff
@@ -0,0 +1,52 @@
...
+created: 2026-04-27
+requires: ["CAIP-2"]
+---
+
+# Namespace for Keeta
+
+[Keeta][keeta-home] is a Delegated Proof of Stake (DPoS) Layer-1 designed for high-throughput asset transfers, real-world asset tokenization, and built-in compliance primitives (KYC, sanctions, ACLs).
+The protocol uses a [Directed Acyclic Graph (DAG)][whitepaper] of per-account block chains rather than a single linearly-ordered ledger, which lets independent accounts publish concurrently and lets the network achieve sub-second finality.
+
+The `keeta` namespace describes both the public Keeta main network and any of its sibling networks -- the test, staging, and dev networks, and any privately-launched **sub-network** (a private instance that runs the same protocol but with isolated transaction visibility and its own validator set).
```
bumblefudge, line 17: the boundaries between networks remains a little ambiguous to me-- by calling them "sub-networks" and referring to the whole as a DAG, i am assuming multiple sub-networks can be "merged" without conflicts to create multi-network views? i would recommend making this more explicit for readers hearing about keeta for the first time and less familiar than me with DAGs.

keeta/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Registering the `keeta` namespace enables standard CAIP-2 chain identifiers for every Keeta network distinguished by their numeric `networkId`.
+Keeta does not run an EVM or similar and has no smart-contract addresses.
+Therefore, CAIP profiles in this namespace focus on identifying networks, accounts (keyed and identifier-style), and tokens, not contracts.
+
+## Governance
+
+Keeta uses Delegated Proof of Stake.
+Token holders elect representatives, which vote on the validity of vote staples (atomic groups of blocks).
+Network-wide policy -- for example, the right to create new tokens or storage accounts -- is expressed as permissions on the [Network Account][accounts], a deterministically-generated identifier account that exists once per network.
```
bumblefudge, line 30: so if a keypair is used on multiple networks, it can have different permissions on different Network Accounts?

keeta/caip2.md
```diff
@@ -0,0 +1,140 @@
...
+
+# CAIP-2
+
+_For context, see the [CAIP-2][] specification._
+
+## Introduction
+
+A Keeta network -- main, test, staging, dev, or any privately-launched network -- is uniquely identified by a single non-negative integer called its `networkId`.
+A privately-launched network is a private instance that runs the same Keeta protocol but with isolated transaction visibility and its own validator set; it shares the key-pair format and address encoding with the public networks but is distinguished from them solely by its `networkId`.
+The same `networkId` is used by the protocol to deterministically derive the network's [Network Account][accounts] and base token address, so two networks with the same `networkId` are by construction the same network.
```
bumblefudge, line 20: two networks with the same networkId are by construction the same network.

but if they have isolated txn visibility, they're not, they're two independent shards of the same network that don't see each other's transactions until somehow merged or bridged, right? this isn't very clearly explained tbh

keeta/caip2.md
```diff
@@ -0,0 +1,140 @@
...
+The CAIP-2 namespace is `keeta`. The CAIP-2 `reference` MUST be the network's `networkId`, rendered as a decimal integer with no leading zeros and no `0x` prefix:
+
+```
+keeta:<networkId>
+```
+
+The `reference` MUST match:
+
+```
+^(0|[1-9][0-9]{0,31})$
```
bumblefudge, line 54: If there is a special case for `networkId`===0, you should explain it in this document.  is it a "public key regardless of network" placeholder for CAIP-10 purposes? if so, explain it in both CAIP-2 and CAIP-10 documents, and explain if it can be used in standalone form (i.e. as a CAIP-2), or just as a CAIP-10.

keeta/caip10.md
```diff
@@ -0,0 +1,190 @@
...
+
+The `<address>` segment MUST match:
+
+```
+^keeta_[a-z2-7]+$
+```
+
+A regex match is necessary but not sufficient for validity as it does not verify the type indicator, checksum, or body length.
```
bumblefudge, line 80:
```suggestion
A regex match is necessary but not sufficient for checking address validity as it does not verify the type indicator, checksum, or body length.
```

Jennie867-5309:
May I just say my stuff always goes to 404.


Jennifer Wages

Jennie867-5309:
I’m novice but I’m trying.


Jennifer Wages

keeta/caip10.md
```diff
@@ -0,0 +1,190 @@
...
+);
+const client = KeetaNet.Client.fromNetwork("test");
+const { info } = await client.getAccountInfo(account);
+```
+
+Implementations SHOULD validate the checksum by round-tripping through `Account.fromPublicKeyString()` before treating the identifier as well-formed.
+
+Implementations SHOULD use the methods of a Keeta SDK (e.g. `account.isToken()`, `account.isStorage()`) to determine the type of an account.
+
+Note that account _existence_ is not required for a CAIP-10 string to be valid -- Keeta accounts come into existence when their first block is published, and a freshly-derived keyed address may be a perfectly valid recipient before any block has ever been published to its chain.
```
bumblefudge, line 107: this might be a good place to explain the `networkId` 0 special case...

keeta/caip10.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+## Additional Considerations
+
+### Security
+
+Keeta addresses encode their algorithm and type in the leading body characters and end in a checksum, so single-character typos and accidental cross-pasting between account types (e.g. sending to a token address instead of a keyed account) are detectable on the client side without a network round-trip.
+Implementations SHOULD perform this check before signing a transaction that references a counterparty CAIP-10.
+
+The same key pair derives to the same `keeta_...` address on every Keeta network, but the `<networkId>` segment of a CAIP-10 string is not redundant: the on-chain _state_ of an account (balances, permissions, certificates, history) is per-network, and applications MUST treat `keeta:21378:keeta_...` and `keeta:1413829460:keeta_...` as different accounts even when the address bodies are identical.
```
bumblefudge, line 172: I'd recommend moving this paragraph to the "Introduction" or other section where fewer readers will miss/skip it, it's a crucial semantic assertion useful to first-timers!

bumblefudge (changes requested):
Mostly minor feedback but `_` is an invalid character for CAIP-10s, and `keeta_` is kind of redundant if `keeta:` is already being prefixed before the network id, so i'd drop the prefix and have a think on how non-account, non-token addresses can be expressed in CAIP-compatible ways!

keeta/README.md
xescure: A Keeta network is one large DAG. That DAG is made of account chains, each a proper chain of blocks belonging to one account. Operations in those blocks can point at other account chains. For example, a `SEND` from `keeta_aa..aa` to `keeta_ab..bb` lives on `a`’s chain and only holds a soft pointer to `b`. The collection of these account chains and their operations is one Keeta network and its ledger.

image

xescure: Subnetworks are networks subordinated to a full network. They operate very similarly, with their own set of validators, except they are forever connected to their supernetwork. Account balances can flow freely and trustlessly between a supernetwork and its subnetworks, but not between unrelated networks.

The proposed spec does not incorporate subnetworks into the CAIP-2 namespace, and only implements full networks. I think this happened because subnetworks are only a planned feature, not something that exists today. Syntax in the form of `keeta:<networkId>-<subnetworkId>` could work to identify a subnetwork. It's TBD if we should do that today or as part of a later expansion when this feature is released.

keeta/README.md
xescure: Yes, exactly. The account corresponding to a keypair can have completely different permissions and balances across networks.

keeta/caip2.md
xescure: Every network is a network with its own validators and own ledger. Networks can be deployed in various contexts. Some of them may be in centralized enterprise environments where some restrictions might be in place. Regardless, these are all just networks (and subnetworks - as described above) and shouldn't be treated any differently by this CAIP standard.

Just doing a quick pass now. Will try to adjust so it's better explained in the spec

xescure:
Ready for your review @bumblefudge - the new version should read better for first time readers, clarifies the ambiguities you pointed out (esp. wrt. subnets and `networkId` 0), and removes the `keeta_` prefix from the CAIP-10 and CAIP-19 specs

xescure:
@lucasrosa90 has created a wallet session protocol that I'm planning to adapt to the relevant CAIPs, but I think it makes sense to keep those as a separate future PR

PR #188, opened 2026-07-12
Add xync namespace (CAIP-2, CAIP-10, CAIP-19 profiles)

mixartemev:
This PR registers the **xync** namespace: Xync Network, a
payments-only Layer-1 blockchain (send / request money in up to 255
protocol-native currencies; no VM). The namespace ships three CAIP
profiles following `_template`:

- **CAIP-2** — human-readable network references declared in genesis
  (`xync:main`, `xync:dev-1`); resolution via any validator's
  `GET /status`, genesis SHA-256 co-signed by the checkpoint quorum.
- **CAIP-10** — accounts are protocol-issued integer indexes
  (`xync:main:518-K7`). Keys are rotatable, so the index is the stable
  identifier; an optional 2-char Crockford-base32 checksum suffix is
  required in user-facing forms (the index space is dense, so typo
  protection matters more than in hash-address namespaces; the
  checksum also binds the CAIP-2 id, catching testnet/mainnet mixups).
- **CAIP-19** — assets are single-byte registry codes
  (`xync:main/cur:0` = the native token XYNC; `cur:1` = issuer-backed
  USD), resolvable via `GET /currencies`.

The network runs a live payment product (XyncPay, Telegram MiniApp)
currently completing progressive decentralization; the reference
implementation and the yellow paper are linked from the profiles.

### DIFF
--- a/xync/caip10.md
+++ b/xync/caip10.md
@@ -2,7 +2,7 @@
 namespace-identifier: xync-caip10
 title: Xync Network - Account ID Specification
 author: Xync Network (@XyncNet)
-discussions-to: TODO-PR-URL
+discussions-to: https://github.com/ChainAgnostic/namespaces/pull/188
 status: Draft
 type: Informational
 created: 2026-07-08
--- a/xync/caip10.md
+++ b/xync/caip10.md
@@ -1,0 +1,112 @@
+---
+namespace-identifier: xync-caip10
+title: Xync Network - Account ID Specification
+author: Xync Network (@XyncNet)
+discussions-to: TODO-PR-URL
+status: Draft
+type: Informational
+created: 2026-07-08
+requires: ["CAIP-2", "CAIP-10"]
+---
+
+# CAIP-10
+
+*For context, see the [CAIP-10][] specification.*
+
+## Introduction
+
+Xync accounts are compact non-negative integer indexes issued
+sequentially by the protocol at registration (account `0` is the
+genesis operator). Public keys are rotatable (`rekey` operation), so
+the stable identifier of an account is its index — NOT a key or a hash
+of one. An optional 2-character checksum suffix protects against typos
+and cross-network confusion; it is REQUIRED in user-facing contexts.
+
+## Specification
+
+### Semantics
+
+The address is the account index in decimal, optionally followed by
+`-CC` where CC is a checksum. Because the index space is dense (most
+small integers are existing accounts), a mistyped address usually hits
+a *real* account — unlike sparse hash-address namespaces. Wallets,
+QR codes and deep links MUST therefore render the checksummed form;
+parsers MUST validate the checksum when present. Exchanges SHOULD
+require the checksummed form for withdrawals.
+
+### Syntax
+
+```
+account_id:        chain_id + ":" + address
+address:           index | index + "-" + checksum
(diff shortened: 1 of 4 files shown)


xync/caip10.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+## Rationale
+
+1. *Index, not pubkey:* keys rotate (`rekey`); an address derived from
+   a key would break on every rotation. The index maps 1:1 to the
+   28-bit address field of the native 128-bit transaction format.
+2. *Checksum-in-address:* analogous in spirit to [EIP-55][] — a
+   canonical plain form and a checksummed display form coexist within
+   one CAIP-10 grammar (`[-.%a-zA-Z0-9]{1,128}` permits both).
```
bumblefudge:
```suggestion
   one CAIP-10 grammar (`[-.%a-zA-Z0-9]{1,12}` permits both).
```
1. 1-9 + `-` + 2, right?  
2. I don't think I understand the `.%` either... 
3. wouldn't it be `[a-hjkm-pq-tx-zA-HJKM-PQ-TX-Z0-9]{1,12}`? probably not the most important detail to stress, just making sure I understand the spec

xync/caip10.md
```diff
@@ -0,0 +1,112 @@
...
+xync:main:518
+
+# genesis operator account
+xync:main:0
+
+# devnet account (same index, DIFFERENT checksum than mainnet)
+xync:dev-1:518-9Q
+```
+
+(Checksum values above are illustrative; compute per the algorithm.)
```
bumblefudge: Haha, it might be useful to actually compute the checksums so that the examples double as test-vectors people can use for their checksum logic 😄

bumblefudge (approved):
Nits above are more clarity suggestions than blockers, interesting work!

xync/caip10.md
mixartemev: `[-.%a-zA-Z0-9]{1,128}` is not the Xync address grammar, it's a verbatim quote of the [account_address production from CAIP-10](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md?plain=1#L48) itself. The only claim the sentence makes is: "our checksummed display form contains a -, and CAIP-10's container grammar already permits that, so both forms are conformant without any extension." The Xync-specific grammar is the one in the Syntax section above.

To your three points:
1. Yes, 12. Index < 2²⁸ = 268,435,456, so at most 9 digits, plus - plus 2 checksum chars → xync:main:268435455-3J.
2. . and % are not ours — they come from CAIP-10, where % is the RFC 3986 §2 percent-escape introducer for namespaces whose addresses need characters outside the allowed set. A Xync address never contains . or %.
3. Almost. Crockford drops I, L, O, U — so V and W stay in, and O is only accepted as an input alias for 0, never emitted. Also the two halves have different alphabets: the index is decimal-only, the checksum is exactly 2 
Crockford chars. Canonical (upper-case) form:
^(0|[1-9][0-9]{0,8})(-[0-9A-HJKMNP-TV-Z]{2})?$

xync/caip10.md
mixartemev: Good call, sure! I will fix this test cases in the next commit

obstropolos (approved):

PR #189, opened 2026-07-15
bip122: add CAIP-19 asset profile for ordinals inscriptions

ordspv:
Adds a CAIP-19 asset profile for the `ordinals` namespace under bip122. Ordinals inscriptions are a major class of non-native Bitcoin assets but have no chain-agnostic asset identifier today; this profile reuses the canonical, self-certifying inscription id (txid + "i" + index) so a reference stays verifiable against Bitcoin consensus data without a trusted indexer. Sat- and inscription-number addressing are explicitly out of scope as trusted-index artifacts. Reference implementation and verification spec: https://github.com/ordspv/ordspv

### DIFF
--- a/bip122/caip19.md
+++ b/bip122/caip19.md
@@ -2,10 +2,11 @@
 namespace-identifier: bip122-caip19
 title: BIP122 Namespace - Assets (Ordinals Inscriptions)
 author: ordspv (https://github.com/ordspv)
+discussions-to: https://github.com/ordinals/ord/discussions/4547
 status: Draft
 type: Standard
 created: 2026-07-11
-requires: ["CAIP-19", "BIP-122"]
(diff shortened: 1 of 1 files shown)


bumblefudge (approved):
Neat!  If I'm understanding this correctly, this only covers the least-controversial, consensus _foundation_ of Inscriptions, right? I know there is some Discourse about incompatible ways of using this stuff, but as far as my limited Bitcoin knowledge extends, the out-of-scope section sounds like this is the neutral bit?

Same question for the spec-verification link (and the analogies to IPFS and verified-fetch were really helpful, thanks for that): there is consensus around this layer, right?

I would still be +1 to merging this even if there were some controvery/non-consensus here, but edit in honest references to that if so!

obstropolos (approved):

PR #191, opened 2026-07-19
Add Neo namespace and CAIP-2 profile

erikzhang:
## Summary

- Add the base `neo` namespace documentation for blockchains implementing the Neo N3 protocol.
- Define a CAIP-2 profile using the canonical unsigned decimal representation of Neo Network Magic.
- Document JSON-RPC resolution, well-known Neo and NeoFS networks, private-network collision behavior, validation cases, and security considerations.

## Motivation

Neo N3 does not currently have a canonical CAIP-2 identifier, which limits interoperability with chain-agnostic wallets, SDKs, payment protocols, and other multi-chain infrastructure.

This proposal was discussed with the Neo community in [neo-project/proposals#238](https://github.com/neo-project/proposals/issues/238). The discussion supported using `neo` as the namespace and the existing Network Magic value as the CAIP-2 reference.

## Specification

The profile defines Neo chain IDs in the following form:

```text
neo:<network-magic>
```

The reference is the Network Magic encoded as a canonical unsigned decimal string. It must:

- match `^(0|[1-9][0-9]{0,9})$`;
- be within the `uint32` range `0..4294967295`;
- contain no sign, base prefix, separator, surrounding whitespace, or leading zeroes.

Clients can resolve the identifier through the standard Neo JSON-RPC `getversion` method by reading `result.protocol.network`.

The profile includes identifiers for these well-known networks:

| Network | CAIP-2 chain ID |
| --- | --- |
| Neo N3 MainNet | `neo:860833102` |
| Neo N3 TestNet T5 | `neo:894710606` |
| NeoFS MainNet | `neo:91414437` |
| NeoFS TestNet | `neo:735783775` |

Because Network Magic is operator-selected and not globally allocated for private networks, the profile explicitly documents collision behavior and operator responsibilities. It also notes that a CAIP-2 identifier obtained from an RPC endpoint is not, by itself, proof of endpoint authenticity.

CAIP-10 account identifiers and CAIP-19 asset identifiers are intentionally left for follow-up proposals.

## Verification

- The `uint32` Network Magic type is defined by [`ProtocolSettings.Network`](https://github.com/neo-project/neo/blob/master-n3/src/Neo/ProtocolSettings.cs).
- The Neo N3 MainNet and TestNet T5 values match the official [`config.json`](https://github.com/neo-project/neo-node/blob/master-n3/src/Neo.CLI/config.json) and [`config.testnet.json`](https://github.com/neo-project/neo-node/blob/master-n3/src/Neo.CLI/config.testnet.json).
- The NeoFS values match the well-known constants in [`neo-go`](https://github.com/nspcc-dev/neo-go/blob/master/pkg/config/netmode/netmode.go).
- The resolution procedure matches the official [`getversion` JSON-RPC documentation](https://docs.neo.org/docs/n3/reference/rpc/getversion.html).
- The CAIP-2 profile includes positive and negative test vectors covering canonical decimal encoding and both `uint32` boundaries.

## Related discussion

- [Neo community discussion: register a CAIP namespace and CAIP-2 profile for Neo N3](https://github.com/neo-project/proposals/issues/238)

### DIFF
--- a/neo/caip2.md
+++ b/neo/caip2.md
@@ -1,0 +1,195 @@
+---
+namespace-identifier: neo-caip2
+title: Neo Namespace - Chains
+author: Erik Zhang (@erikzhang)
+discussions-to: https://github.com/neo-project/proposals/issues/238
+status: Draft
+type: Standard
+created: 2026-07-19
+requires: CAIP-2
+---
+
+# CAIP-2
+
+*For context, see the [CAIP-2][] specification.*
+
+## Introduction
+
+Every Neo N3 node is configured with an unsigned 32-bit Network Magic value.
+The `neo` CAIP-2 profile identifies a blockchain by encoding that value as a
+canonical unsigned decimal string.
+
+## Specification
+
+### Semantics
+
+A Neo chain ID consists of the literal namespace `neo`, a colon, and a
+reference containing the blockchain's Network Magic.
+Two blockchains configured with the same Network Magic have the same CAIP-2
+chain ID by definition.
+
+### Syntax
+
+```text
+chain_id:  "neo:" + reference
+namespace: neo
+reference: 0 | [1-9][0-9]{0,9}
+```
+
+The reference must match the following regular expression:
+
+```regex
+^(0|[1-9][0-9]{0,9})$
+```
+
+The parsed value must additionally be in the unsigned 32-bit integer range
+`0` through `4294967295`, inclusive.
+The regular expression alone is not sufficient to enforce the upper bound.
+
+The reference uses ASCII decimal digits without a sign, base prefix,
+separators, surrounding whitespace, or leading zeroes.
+The single value zero is encoded as `0`.
+The namespace is lowercase and CAIP-2 identifiers are case-sensitive.
+
+### Resolution Mechanics
+
+To resolve the chain ID reported by a Neo N3 JSON-RPC endpoint, call the
+[`getversion`][] method:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "getversion",
+  "params": [],
+  "id": 1
+}
+```
+
+The response contains the Network Magic in `result.protocol.network`.
+The following is an abbreviated response from a MainNet node:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "protocol": {
+      "network": 860833102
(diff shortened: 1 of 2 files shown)


neo/caip2.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+The parsed value must additionally be in the unsigned 32-bit integer range
+`0` through `4294967295`, inclusive.
+The regular expression alone is not sufficient to enforce the upper bound.
+
+The reference uses ASCII decimal digits without a sign, base prefix,
+separators, surrounding whitespace, or leading zeroes.
+The single value zero is encoded as `0`.
+The namespace is lowercase and CAIP-2 identifiers are case-sensitive.
```
bumblefudge: Wait, why does it need to be lowercase and/or case-sensitive if the only valid values are integers? Are you referring to CAIP-10s here?  I'd remove this line (or move it to a future CAIP-10/19/etc), because integers have no case!

```suggestion
```

neo/caip2.md
```diff
@@ -0,0 +1,195 @@
...
+^(0|[1-9][0-9]{0,9})$
+```
+
+The parsed value must additionally be in the unsigned 32-bit integer range
+`0` through `4294967295`, inclusive.
+The regular expression alone is not sufficient to enforce the upper bound.
+
+The reference uses ASCII decimal digits without a sign, base prefix,
+separators, surrounding whitespace, or leading zeroes.
+The single value zero is encoded as `0`.
```
bumblefudge, line 51: I would explain [here] what `0` means-- is that a valid Network Magic, or is it special-case (like "offchain" or "placeholder" or "dead network")?  I would also explain what a CAIP-10 with a `0` chainId means in a future CAIP-10 profile, as these semantics vary widely across L1s/VMs and multichain developers need this spelled out!

bumblefudge (approved):
Seems fine, modulo the corrections

neo/caip2.md
erikzhang: Here the `namesapce` is "neo" which must be lowercase. The `reference`s are integers.

neo/caip2.md
erikzhang: Thanks. I’ve clarified that 0 is a valid reference and has no special meaning.

obstropolos (approved):

PR #224, opened 2026-09-14
Add antelope CAIP-19 asset profile (EOS, WAX, Telos)

okwme:
Adds a CAIP-19 asset profile to the existing `antelope` namespace. The CAIP-2 profile is already ratified and is not touched.

### Who wrote this

**I do not represent the antelope ecosystem.** I maintain [FAR](https://github.com/0xcounting/FAR), an open cross-reference between CAIP-19, CoinGecko and ISO 24165 (DTI) identifiers, which needs a way to name **6 assets** on this chain. CASA's README invites profiles added to an existing namespace, so this documents how assets are addressed rather than claiming to speak for anyone. Offered for the antelope community to correct, adopt or replace.

### Asset namespaces

`slip44, token, nft`

### What this profile establishes

**Contract first, then symbol, separated by a single `-`.** Account names are `[.12345a-z]` and symbol codes are `[A-Z]`, so the two halves have **disjoint character classes** and a reference self-validates offline with no escaping. Symbol-first with a dotted account name would have needed a parse-from-the-right rule.

**The safety case is live and is the reason the contract must be in the identifier.** On EOS mainnet the symbol `EOS` is issued by `eosio.token`, by `iouiouiou123` (100,000,000.0000 EOS), **and** by `eosiotokenis` at *zero precision* — a consumer matching on symbol alone misreports that last one by a factor of 10⁴. `USDT` is issued by four different contracts.

Precision is deliberately excluded from the reference: an Antelope symbol is `4,EOS`, `,` is not in the charset, and precision is not identity.

Every example was verified against a live node or explorer; anything that could not be verified is marked as such in the document rather than asserted. `discussions-to` points at https://github.com/0xcounting/FAR/issues/1.


Jennie867-5309:
Everyone is using my crypto and Apple ID. I can’t get anything to work. I
had a heart attack no food no medicine no bills paid please help me


Jennifer Wages

